PR #51, opened 2013-04-21
Add whitelist option for list-filter to prevent arbitrary tag generation

ignatz:
Some of my mail contains weird list-id headers (e.g. airline mails), tricking afew into creating weird tags. To prevent this from happening i added a 'whitelist' option to the 'ListMailsFilter'. If empty (default), afew behaves just the same. With a non empty whitelist any list mail still gets the 'lists' tag, but only those mails with a matching list-id get the extra 'list-id' tag.

### DIFF
--- a/afew/filters/ListMailsFilter.py
+++ b/afew/filters/ListMailsFilter.py
@@ -24,12 +24,21 @@ from ..Filter import Filter, register_filter
 @register_filter
 class ListMailsFilter(Filter):
     message = 'Tagging mailing list posts'
-    query = 'NOT tag:lists'
+    list_tag = 'lists'
+    query = 'NOT tag:%s' % list_tag
+
+    def __init__(self, whitelist=''):
+        super(ListMailsFilter, self).__init__()
+        self.whitelist = set(whitelist.split())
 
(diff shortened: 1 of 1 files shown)


acowley:
The `list/` namespace is definitely helpful, but the interaction with other clients is not always ideal (e.g. the iOS Mail app doesn't show the `list` namespace collapsed). Granted, that is not `afew`'s problem, and one can write a sequence of `[HeaderMatchingFilter.X]` sections to do manual tagging based on list-id. I'm just chiming in on the off chance this ever gets revisited because the feature implemented here is exactly what I was looking for.

GuillaumeSeren:
Hey @acowley ,
Nice digging here ;)

I aggree on the 'accepted-list of list-id can be usefull' as I have plenty of not usefull list tags that I need to cleanup,
every once in a while, most of them come from spam / commercial stuff that use a MailingList behind the scene."

There already a '_blacklist' array in the BaseFilter (not a fan of the name, could be nice to upgrade those),
so there could be a '_allowlist' in there too, making available to use in all filters, can be a nice PR if you interrested.

[BaseFilter](https://github.com/afewmail/afew/blob/master/afew/filters/BaseFilter.py#L8)

PR #73, opened 2014-02-20
FolderNameFilter: add folder_lowercases

tsgates:


GuillaumeSeren:
Hey @tsgates 
That seems promising but can you develop a bit the use case ?

flokli:
I think the usecase is clear. However, I'd go even further, and pipe all folder names through [python-slugify](https://github.com/un33k/python-slugify) by default, to get sane tags.

This could be the default case, with an option to disable it if needed.

GuillaumeSeren:
@flokli Maybe you want to upgrade the patch because the owner has left the PR for 3+ years ?

flokli:
@GuillaumeSeren No need to rebase, as the patch still applies cleanly. Let's wait if we hear something from @tsgates about my suggestion.

GuillaumeSeren:
@flokli Ok !

GuillaumeSeren:
Hey,
let's merge this, the patch is old and the user is not responding.

@flokli If you want to change the behaviour to use 'python-slugify', it could be a later upgrade.

PR #78, opened 2014-04-03
filter, test: sort filters before asserting

gauteh:
existing filters and hardcoded list doesn't match because
they do not have the same order. sort both before testing.

the deprecated assertEquals have been replaced with assertEqual.

tested on python 3.4.


GuillaumeSeren:
LGTM 👍

GuillaumeSeren:
Merged ! thanks @gauteh 👍

PR #87, opened 2014-07-18
MailMover : Use folder replacement and regex in settings.

phlb:
This patch allows to match folder by regex in the mailmover rules and to replace the string '{folder}' by the current folder in the notmuch query. This avoid to duplicate the same rule on each folder. 

For exemple : 

```
[MailMover]
folders = .Lists.firstlist .Lists.secondlist

\.Lists\..* =
    'tag:spam':.Junk
    '(NOT tag:spam) AND (NOT folder:.Archive{folder}) AND tag:archives':.Archive{folder}

...
```

### DIFF
--- a/afew/Settings.py
+++ b/afew/Settings.py
@@ -79,16 +79,20 @@ def get_mail_move_rules():
     rule_pattern = re.compile(r"'(.+?)':(\S+)")
     if settings.has_option(mail_mover_section, 'folders'):
         all_rules = collections.OrderedDict()
+        all_folder_regex = list()
+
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hey @phlb 
As a big user of this feature I think if we can have a way to not copy/paste every rules,
on each account would be neat, seems similar to #88 to me.

As it is 'a big' feature of afew we need to take care not breaking it.

GuillaumeSeren:
Due to lack of response from the user, let's close this

PR #88, opened 2014-08-18
Change semnatics of MailMover

mnick:
Hi,
I would like to propose a slight change in the semantics of `MailMover` rules. Right now, `MailMover` requires a fixed source folder for which a query applies. However, I find this very inconvenient for rules like 

```
Inbox = 'tag:deleted':Trash
```

since it requires to manually enter this rule for each maildir folder separately. This is especially inconvenient since notmuch also doesn't sync the deleted flag anymore.

The proposed patch changes the semantics of MailMover, such that `Trash =` in

```
Trash = 'tag:deleted AND NOT folder:Trash':Trash
```

only specifies the ID of the rule, but not the source folder. If a source folder is required, this could be added to the query separately, e.g.

```
Trash = 'tag:deleted AND folder:Inbox':Trash
```

I find this much more flexible then the previous semantics of `MailMover`.

However, I understand that this might break the setup for all users that currently rely on the old semantics. To fix this, the proposed `MailMover` could be added as an additional mover, e.g. `QueryBasedMailMover`.

Cheers,
Max

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -87,7 +87,7 @@ class MailMover(Database):
         # update notmuch database
         logging.info("updating database")
         if not self.dry_run:
-            self.__update_db(maildir)
+            self.__update_db(rule_id)
         else:
             logging.info("Would update database")
 
@@ -96,7 +96,7 @@ class MailMover(Database):
     # private:
     #
 
-    def __update_db(self, maildir):
+    def __update_db(self, rule_id):
         '''
         Update the database after mail files have been moved in the filesystem.
         '''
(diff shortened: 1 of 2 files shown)


GuillaumeSeren:
Hey @mnick 
I come late to the party, but I try to help moving the project forward.
As big user of that feature I fear the breaking change, but I agree on the problem which is if I understand correctly, 'Optimise the rules of the mail mover'.

In my case I get something like
```
mail1/INBOX = 'tag:spam':mail1/spam 'tag:sent':mail1/sent
mail2/INBOX = 'tag:spam':mail2/spam 'tag:sent':mail2/sent
```

And that is the same for each of my account, so I could use a patch that change the behaviour to became more generic, and keep a way to specify the folder if needed.

GuillaumeSeren:
Due to lack of response from the user, let's close this.

PR #91, opened 2014-10-17
Behave well when using mbsync to sync maildirs

larkery:
I use [mbsync](http://isync.sourceforge.net/mbsync.html) to sync my maildir with remote IMAP, and it has an odd behaviour where moving a mail from one maildir to another without changing its name breaks the sync state because the UID for the message is derived from the filename alone; it seems like this is a limitation of mbsync which is otherwise a good bit of software. 

These changes add an option `rename` to `MailMover` which makes it concoct a new filename when it moves mail to another maildir. A similar workaround exists in e.g. [mu4e](https://code.google.com/p/mu0/issues/detail?id=76).

I'm guessing most afew users are either using offlineimap or some other tool for producing their maildir which is why nobody else has encountered this problem before.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -34,7 +34,7 @@ class MailMover(Database):
     '''
 
 
-    def __init__(self, max_age=0, dry_run=False):
+    def __init__(self, max_age=0, rename = False, dry_run=False):
         super(MailMover, self).__init__()
         self.db = notmuch.Database(self.db_path)
         self.query = 'folder:{folder} AND {subquery}'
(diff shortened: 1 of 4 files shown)


philc:
Any update on this? I'm also using mbsync.

varac:
@larkery thx for the pull req, with "rename = true" it works for my now with mbsync.
Just a hint: You should consider updating the documentation as well on the next pull req ;)

larkery:
Happy to help; it's been working for me for a while now, so I think it is probably trustworthy. However, someone who has actually read the maildir spec might have a different opinion about what I'm doing, so caveat emptor.

hugoroy:
I've added `rename = True` in afew's config for mail moving but still get the UID issue with mbsync. Any idea?

PR #110, opened 2015-07-24
Don't interfere with the notmuch unread tag.

hlieberman:
Right now, afew tramples the notmuch unread tag, interfering with notmuch-el.

This patch will change that behavior to not modify the unread tag.

### DIFF
--- a/afew/NotmuchSettings.py
+++ b/afew/NotmuchSettings.py
@@ -30,7 +30,12 @@ def read_notmuch_settings(path = None):
     notmuch_settings.readfp(open(path))
 
 def get_notmuch_new_tags():
-    return notmuch_settings.get_list('new', 'tags')
+    tags = notmuch_settings.get_list('new', 'tags')
+    try:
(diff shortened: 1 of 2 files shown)


babilen:
This would be very useful to have. For now I am using the following to approximate this behaviour:

```
[Filter.0]
message = 'Retags all messages not tagged as junk or killed as inbox'
tags = +inbox;-new
tags_blacklist = killed;spam
```

GuillaumeSeren:
Hey @hlieberman 
I didn't notice this behaviour until reading your PR, to me it might more be something that the user can configure, than just changing the default for everybody.

GuillaumeSeren:
Hey,
this is not moving and get no negative feedback, let's merge it.

flokli:
This basically triggered #158.

GuillaumeSeren:
@flokli Yes it is true,
if it is a problem we can revert it, but I've only tested them on python-2, now I'll test on both python-2 and python-3 🤖

PR #111, opened 2015-08-04
[utils] avoid stripping all lines when no signature is found

adimascio:
When no signature is found, siglines equals to 0 and lines[:-0] will strip off all lines rather than keeping content intact.

### DIFF
--- a/afew/tests/test_utils.py
+++ b/afew/tests/test_utils.py
@@ -1,0 +1,40 @@
+#
+import unittest
+
+from afew import utils
+
+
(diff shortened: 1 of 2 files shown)


GuillaumeSeren:
Hey @adimascio 
This look good to me, and there is a test with the patch, for starters can you rebase your branch on master ?

kyrias:
There are no merge conflicts, so doesn't seem to be much point in rebasing?

GuillaumeSeren:
Yes this is right @kyrias 
let's merge it

PR #115, opened 2015-12-26
Allow quoted folder names in MoveMail settings

khinsen:
Important for folder names containing spaces.

### DIFF
--- a/afew/Settings.py
+++ b/afew/Settings.py
@@ -77,7 +77,7 @@ def get_filter_chain(database):
     return filter_chain
 
 def get_mail_move_rules():
-    rule_pattern = re.compile(r"'(.+?)':(\S+)")
+    rule_pattern = re.compile(r"'(.+?)':((?P<quote>['\"])(.*?)(?P=quote)|\S+)")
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
I am not really sure about the use case exactly, one of you imap folder is quoted right ?

Why don't you translate it when you synchronise to get 'normal' dir name,
like in offlineimap I do something like:
```
nametrans = lambda folder: {
    'archive'  : '[Gmail]/Tous les messages',
    'drafts'   : '[Gmail]/Brouillons',
    'trash'    : '[Gmail]/Corbeille',
    'important': '[Gmail]/Important',
    'sent'     : '[Gmail]/Messages envoy&AOk-s',
    'spam'     : '[Gmail]/Spam',
    'starred'  : '[Gmail]/Suivis'
```

khinsen:
I use both notmuch and traditional IMAP clients on the same mail accounts. And since I have lots of IMAP folders, it is important for me that their names are the same everywhere.

GuillaumeSeren:
Hey @khinsen 
it is not my use-case but it can be useful and the patch LGTM,
sorry for the delay.

Issue #24, opened 2011-12-24
use argparse for command line parsing


teythoon:

lucc:
close?

GuillaumeSeren:
Closing

PR #117, opened 2016-04-04
Switch from optparse to argparse.

lucc:
This is a PR for #24.


GuillaumeSeren:
Hi @lucc ,
as this solve a problem reported by author I think we should take it.

It is probably needed to add the new dependency in the 'setup.py' install file ?

lucc:
argparse in the standard lib since 2.7 so it is not an external dependency we need to install.

GuillaumeSeren:
@lucc LGTM but let's give some time to see if other maintainer want to add something

GuillaumeSeren:
Merged ! thanks @lucc 👍

PR #119, opened 2016-04-18
Implement archiving mails into folders

aperezdc:
This adds an operation which can be used to file (archive) messages into folders. The new `MailArchiver` class is actually quite similar to `MailMover`, with the (important) addition that the destination folder name is expanded for variables. Ccurrently only the message date is provided for expansion, which allows to configure it like this:

```
[MailArchiver]
folders = INBOX
INBOX = 'not tag:flagged and not tag:inbox and not tag:unread':Archives/{date.year}/{date.year}-{date.month:0>2d}
```

That will move seen and unflagged messages from `INBOX` into an archives subfolder which contains the date and month of the message in its name. This
creates a folder structure like this:

```
Archives/
  2015/
    2015-11/
    2015-12/
  2016/
    2016-01/
    2016-02/
    ...
```

### DIFF
--- a/afew/MailArchiver.py
+++ b/afew/MailArchiver.py
@@ -1,0 +1,118 @@
+# coding=utf-8
+
+#
+# Copyright (c) dtk <dtk@gmx.de>
+#
+# Permission to use, copy, modify, and distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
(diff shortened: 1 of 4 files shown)


aperezdc:
One thing I have noticed after implementing `MailArchiver` is that the final result is quite similar to `MailMover`. As an alternative to this patch, I could make a different one which modifies `MailMover` to allow expanding variables in the name of the destination folder. WDYT?

aperezdc:
Could you please take a look at this PR? I have been maintaining my locally patched version of `afew`, but I pretty much would prefer for this to be available upstream — I can imagine how this would be useful for many more people :wink:

GuillaumeSeren:
Hey @aperezdc 
I think it could be useful, and as it is only adding a new feature it should be safe to pull.

> One thing I have noticed after implementing MailArchiver is that the final result is quite similar to MailMover. As an alternative to this patch, I could make a different one which modifies MailMover to allow expanding variables in the name of the destination folder. WDYT?

And I agree it could be nice, to have variable expansion for the MailMover as well.

aperezdc:
@GuillaumeSeren: Thanks for commenting. I'll try to rework the PR to add the functionality in `MailMover` when I have a few spare cycles :wink:

GuillaumeSeren:
Hey @aperezdc 🍺 
There at least 3 open PR on the subject (MailMover),
so please let is really small because I don't want to break it.

I don't really mean, to add all the features in MailMover but try to keep a common base,
where it is possible and avoid code duplication (as you said the 2 features are similar).

aperezdc:
@GuillaumeSeren: I will see if I can factor our the common functionality, and make `MailArchiver` a subclass of `MailMover`. That should keep the changes needed in `MailMover` down to a minimum.

GuillaumeSeren:
@aperezdc Yes exactly that would be great ! 🍺

GuillaumeSeren:
Hey @aperezdc ,
is there any news on this PR ?

PR #118, opened 2016-04-18
Log user filters instead of printing

ankurdave:

PR #120, opened 2016-07-31
Import user filters silently (INFO priority)

wavexx:
There's a stray print() when importing user filters. Use logging.info instead.


wavexx:
Sorry, didn't notice there is already the same PR #118

PR #128, opened 2017-02-09
Add a __main__.py file

lucc:
This makes it easier to run the program directly from the git folder without
prior installation, which is especially useful during development and for
quick tests.

It can be used with pythons -m switch:

PYTHONPATH=. python2 -m afew --help
PYTHONPATH=. python3 -m afew --tag --new --dry-run

### DIFF
--- a/afew/__main__.py
+++ b/afew/__main__.py
@@ -1,0 +1,21 @@
+# coding=utf-8
+from __future__ import print_function, absolute_import, unicode_literals
+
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
@lucc I am not a python expert but I know it is convenient to have one.

GuillaumeSeren:
Merged thanks @lucc 👍

PR #138, opened 2017-04-10
Fix #136 learn/update not usable due to argparse migration

### LINKED ISSUE #136: Learn/update not usable since migration to argparse
As far as I can tell, the migration to argparse somehow made the options to -l and -u vanish.
Symptom:

    % afew --learn spam tag:spam

    Traceback (most recent call last):
      File ".local/bin/afew", line 11, in <module>
        load_entry_point('afew==1.0.0', 'console_scripts', 'afew')()
      File "/home/jan/.local/lib64/python3.4/site-packages/afew-1.0.0-py3.4.egg/afew/commands.py", line 205, in main
        inner_main(args, database, query_string)
      File "/home/jan/.local/lib64/python3.4/site-packages/afew-1.0.0-py3.4.egg/afew/main.py", line 49, in main
        database.mail_bodies_matching(query_string)
      File "/home/jan/.local/lib64/python3.4/site-packages/afew-1.0.0-py3.4.egg/afew/DBACL.py", line 78, in learn
        process = self._call_dbacl(['-l', self.get_category_path(category)])
      File "/home/jan/.local/lib64/python3.4/site-packages/afew-1.0.0-py3.4.egg/afew/DBACL.py", line 75, in get_category_path
        return os.path.join(self.database_directory, category.replace('/', '_'))
    AttributeError: 'bool' object has no attribute 'replace'

This is with the latest git revision.

opensamba:
Hi,
I double checked the prev code and seems after the argparse migration the learn option was not configured correctly.

I tested and seems it's working fine for me.

please have a look.

### DIFF
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -50,7 +50,7 @@ action_group.add_argument(
     help='continuously monitor the mailbox for new files'
 )
 action_group.add_argument(
-    '-l', '--learn', action='store_true',
+    '-l', '--learn', action='store',
     help='train the category with the messages matching the given query'
 )
 action_group.add_argument(


GuillaumeSeren:
@opensamba Hey I look at it ASAP !

Thank for your feedback !

afew/commands.py
```diff
@@ -50,7 +50,7 @@
     help='continuously monitor the mailbox for new files'
 )
 action_group.add_argument(
-    '-l', '--learn', action='store_true',
+    '-l', '--learn', default=False,
```
andir: You should not drop `action='store_true'` since argparse will expect an argument in the non-default case.

opensamba:
sent another push, 
what about now?

flokli:
I'm a bit confused - can you squash those commits together for better readability?

It seems like you only set the `default` argument to `False`, which should already be `False`, if `store_true` action is used and the argument is not passed - at least according to the example in the [documentation](https://docs.python.org/dev/library/argparse.html#action), and [redundant](http://stackoverflow.com/questions/8259001/python-argparse-command-line-flags-without-arguments#comment10163905_8259080).

opensamba:
Hi,
sorry if messing with multiple git commit, still figure out how it works..

ok, now some info: It's not clear for me too how it should be.

in the commit [before the migration to argparse](https://github.com/opensamba/afew/blob/40aa6cf8098bc6392ca28266f16b5c5238c6e280/afew/commands.py) the code was:
`action_group.add_option(
    '-l', '--learn', default=False,
    help='train the category with the messages matching the given query'
)`

so I guess the option should be  default=False, without the store_true.

in this way we can force an argument. if not we get the error we have now, because opt.learn=False|True which is bool, while it should be a string

andir:
Even if it might not be strictly required as @flokli said I think the latest version you submitted makes the desired behaviour clear.

Regarding squashing:
- We can do that while merging your code
OR
- You squash them together using `git rebase -i HEAD~2`. Then change `pick` to `squash` on the latest commit. You'll be asked to update your commit message (since those two commit will the ben one). Finally you'll have to push with `--force` since git can't fast-forward anymore.

opensamba:
Hi @andir ,
thanks for your support with this!

I did the rebase + squash, then I edit the commit and pushed with `--force`
please let me know if now is it ok.

Bye

andir:
Now we've 3 commits :-)

Do you want to try again? Otherwise I can squash that while merging.

opensamba:
haha, yes, let me try..

opensamba:
For what I see the learn opt can't contain a bool, because if we have a bool then we have the issue #136 

So I think we should have just an action="store" , for that I edit my commit and push it again

GuillaumeSeren:
Sorry but I was away last week so I did not investigate on the issue.

@opensamba Do you mind to squash the 3 commits into 1 ?

Also please add a small description of the use case / fix in your commit.

opensamba:
done.
what about the CI, is it normal is failing at every push ?

toogley:
@opensamba yes it does at the moment. travis is enabled but no .travis.yml is in the repository.

GuillaumeSeren:
Let's merge this thank you @opensamba !

PR #140, opened 2017-04-25
options.learn can be None or False, don't use an `is' check

twopir:


GuillaumeSeren:
@twopir LGTM let's merge it

PR #142, opened 2017-05-20
Add version option

lucc:
The version number is extracted from the git tag.


GuillaumeSeren:
Yes I agree ! 🍻 

Thank's @lucc

PR #74, opened 2014-03-01
FolderNameFilter handles messages in multiple folders.

daewok:
The FolderNameFilter used to look only at a message's primary folder (message.get_filename()) when deciding which tags to apply. This change allows the filter to look at all folders that a message exists in when applying the tags.

This was done by retrieving all filenames using message.get_filenames() and adding the appropriate processing to handle the list of file names.

This change would be especially useful for those using GMail to add tags to their messages.

PR #145, opened 2017-05-29
FolderNameFilter now looks at all folders that a message is in.

GuillaumeSeren:
Switched from message.get_filename() to message.get_filenames() and added appropriate processing to handle the list of file names.

This is a rebased version of the #74

### DIFF
--- a/afew/filters/FolderNameFilter.py
+++ b/afew/filters/FolderNameFilter.py
@@ -39,9 +39,18 @@ class FolderNameFilter(Filter):
 
 
     def handle_message(self, message):
-        maildirs = re.match(self.__filename_pattern, message.get_filename())
+        # Find all the dirs in the mail directory that this message
+        # belongs to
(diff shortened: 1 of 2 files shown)


GuillaumeSeren:
I have rebased the branch but not tested it intensively so I need some help on that 🍺

varac:
I used this branch already for a while now and it works for me. But I wouldn't call this extensive testing.

GuillaumeSeren:
@varac Thank you for the feedback 🍺 
I will let the patch wait for a bit more reviews

flokli:
@GuillaumeSeren: I think this might need a bit usage documentation, and we probably also should start adding some tests…

GuillaumeSeren:
Hey @flokli ,
I have added a bit of documentation in the branch, please have a look at it.

Also, I have seen the #152, that was more or less the same thing so I think,
we should merge it, then upgrade it as needed.

GuillaumeSeren:
Ok let's merge it.

PR #146, opened 2017-06-05
If decode fails but email is base64, use regex to get base64 data

kyrias:
Some mailinglists just slam their signature at the end of the email
whether it's base64 encoded or not, leading to afew throwing exceptions
currently, because get_payload doesn't handle that case.

Instead we can try to grab the base64 part out of the message with a
regex and then decode that ourselves.

---

This feels sort of dirty, but it's necessary for quite a few of my emails, and I can't think of a better way to do it off the top of my head.

### DIFF
--- a/afew/utils.py
+++ b/afew/utils.py
@@ -21,6 +21,7 @@ import codecs
 import re
 import sys
 import email
+import base64
 from datetime import datetime
 
 signature_line_re = re.compile(r'^((--)|(__)|(==)|(\*\*)|(##))')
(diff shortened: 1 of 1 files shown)


flokli:
Thanks a lot for your PR!

As we now have a working Travis CI, can you also add a test trying to parse such a message?

kyrias:
Got a bit short on time now, but I will definitely look into it!

kyrias:
Hmm, I don't seem to currently have any emails that cause this issue anymore.  I suspect that they were somehow corrupted on disk, which was fixed when I did a clean sync again.  I'm going to close this PR for now, and then I'll open it again if I can reproduce it.

PR #147, opened 2017-06-10
Make MailMover an ABC and have FolderMailMover as a subclass

kyrias:
This is as a preparation for implementing a version of #88 as an option.  It makes `MailMover` an Abstract Base Class which means that anything deriving from it has to implement the proper methods to be able to instantiate them at all.

For an idea of how the end result will work, this PR adds a configuration option called `global.mail_mover_kind` which defaults to `folder`, which is the current `MailMover` behaviour.  The follow-up PR with the functionality from #88 will add `query` as an option to `global.mail_mover_kind` which will read rules from the `QueryMailMover` configuration section and apply them as in #88.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -25,19 +25,19 @@ from subprocess import check_call, CalledProcessError
 from .Database import Database
 from .utils import get_message_summary
 from datetime import date, datetime, timedelta
+from abc import ABCMeta, abstractmethod
 import uuid
 
 
 class MailMover(Database):
+    __metaclass__ = ABCMeta
     '''
     Move mail files matching a given notmuch query into a target maildir folder.
     '''
-
-
-    def __init__(self, max_age=0, rename = False, dry_run=False):
+    def __init__(self, max_age=0, rename=False, dry_run=False, query=None):
         super(MailMover, self).__init__()
         self.db = notmuch.Database(self.db_path)
-        self.query = 'folder:{folder} AND {subquery}'
+        self.query = query
         if max_age:
             days = timedelta(int(max_age))
             start = date.today() - days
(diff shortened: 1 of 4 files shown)


kyrias:
ping?

GuillaumeSeren:
Hello @kyrias 
Thank you for the work on the mail mover it is a very important feature to me !

I have checked the patch, to me it seems ok, but it would be better with a test don't you agree ?

flokli:
I also really like this, however, I'm a bit tempted to merge this as long as we don't have at least some tests for afew's main functions.

GuillaumeSeren:
Hey @kyrias ,
it would e neat to get this merged, first it need to be rebased, then adding a test,
I understand testing such a case is not easy, maybe someone can help write the test.

GuillaumeSeren:
Hey,
should we close this PR ? (I think yes)

Because owner is not active and MailMover is a big feature for afew,
we don't want to break it.

flokli:
Well, as written before, with a proper test suite we don't have to be afraid of merging this. I'd keep this open until we have one.

GuillaumeSeren:
Yes @flokli I was expecting this from you so I did not close it but if the owner will not upgrade the PR we finally need to create a new one with the tests

afew/MailMover.py
```diff
@@ -25,19 +25,19 @@
 from .Database import Database
 from .utils import get_message_summary
 from datetime import date, datetime, timedelta
+from abc import ABCMeta, abstractmethod
```
andir: please sort import alphabetically

afew/MailMover.py
```diff
@@ -47,7 +47,47 @@ def __init__(self, max_age=0, rename = False, dry_run=False):
         self.dry_run = dry_run
         self.rename = rename
 
-    def get_new_name(self, fname, destination):
+    @abstractmethod
```
andir: Shouldn't the base class (ignoring the abstract here) provide a generic moving implementation?
Moving mails is the tricky part in this feature. We shouldn't encourage duplicating that effort.

GuillaumeSeren:
Hello, @kyrias 
As this PR is not active anymore, I am closing it.
Feel free to reopen if needed with upgrades/replys.

PR #150, opened 2017-07-12
MailMover: Only run 'notmuch new' if any mails were actually moved

tohojo:
When running `afew --move`, the notmuch database is always updated, even if not
files were actually moved. Since this takes a non-trivial amount of time, the
time to run afew can actually be dominated by the database update, which is
annoying when putting afew into an automated 'check new mails' hook.

This adds a small optimisation where the database is only updated if any
messages were actually moved. This makes running `afew --move` substantially
faster in the common case were no messages need to be moved.

I realise this pull requests conflicts with some of the other open pull requests
that also touch MailMover; but since this one is fairly small, it should be
simple to incorporate in those as well :)

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -65,6 +65,7 @@ class MailMover(Database):
         # identify and move messages
         logging.info("checking mails in '{}'".format(maildir))
         to_delete_fnames = []
+        moved = False
         for query in rules.keys():
             destination = '{}/{}/cur/'.format(self.db_path, rules[query])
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hey @tohojo thank for sharing this patch,
to me it is fine and as I heavily use this option saving to update the database if no mail was moved can save a bit of cpu / ssd.

I let other maintainer have a look but I think this could be merged.

GuillaumeSeren:
@flokli Can you have a look ?

flokli:
Merged, thanks!

PR #152, opened 2017-07-20
FolderNameFilter: support messages that exist in multiple locations

wyleyr:
In handle_message, tag a message on the basis of the folder names
in *all* the filesystem paths for that message.

This improves support in particular for GMail, where the same message
may be stored in several different directories.  Notmuch knows to
group these together, and notmuch.message.Message.get_filenames()
returns all the paths for the given message.  This change simply runs
the previous code in a loop over these paths.

### DIFF
--- a/afew/filters/FolderNameFilter.py
+++ b/afew/filters/FolderNameFilter.py
@@ -39,21 +39,22 @@ class FolderNameFilter(Filter):
 
 
     def handle_message(self, message):
-        maildirs = re.match(self.__filename_pattern, message.get_filename())
-        if maildirs:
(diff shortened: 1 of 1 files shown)


wyleyr:
Oh shoot, I see that a very similar PR is already open: #145.  My patch takes a different (slightly more naive) approach. Closing...sorry for the noise.

GuillaumeSeren:
Hey @wyleyr , 🍺 
thank you for the PR, I rebased #74 into #145 (on latest master), 
but lack a bit on reviewers if want you can have a look and add your feedback.

Issue #156, opened 2017-08-02
Error installing afew through pip on macOS


svmhdvn:
I followed your installation guide in the README, but I get an error when I try to install `afew` with `pip`:

```
(.venv) bash-4.4$ pip install afew
Collecting afew
  Using cached afew-1.1.0.tar.gz
Collecting notmuch (from afew)
  Could not find a version that satisfies the requirement notmuch (from afew) (from versions: )
No matching distribution found for notmuch (from afew)
```

These are the versions of the relevant tools:
```
(.venv) bash-4.4$ pip --version
pip 9.0.1 from /Users/siva/.venv/lib/python3.6/site-packages (python 3.6)
(.venv) bash-4.4$ python --version
Python 3.6.2
(.venv) bash-4.4$ notmuch --version
notmuch 0.25
```

I installed notmuch through homebrew. Is there anything I'm missing that I need to install `afew` using `pip`?

flokli:
It's looks pretty much like the notmuch python bindings are missing. Can you install those via homebrew, too?

svmhdvn:
The notmuch brew formula didn't install python3 bindings by default, it only installed the python2 ones. I reinstalled with `brew install --with-python3 notmuch` and it worked! Thanks for the help.

flokli:
No problem! Would you mind adding a section to the README, which explains installation via homebrew?

svmhdvn:
Added, hopefully that's enough :)

PR #157, opened 2017-08-03
Add info about installing notmuch with homebrew

svmhdvn:
Ran into #156, this would help save time for people installing notmuch with homebrew.


GuillaumeSeren:
Hey @svmhdvn ,
Thank you for the patch !

Issue #158, opened 2017-08-05
AttributeError: 'filter' object has no attribute 'remove'


mkaito:
This has been happening for about a week now. No changes made to config when it happened.

```
Traceback (most recent call last):
  File "/usr/bin/afew", line 11, in <module>
    load_entry_point('afew==1.1.1.dev15+g7f942f3', 'console_scripts', 'afew')()
  File "/usr/lib/python3.6/site-packages/afew/commands.py", line 131, in main
    query_string = get_notmuch_new_query()
  File "/usr/lib/python3.6/site-packages/afew/NotmuchSettings.py", line 41, in get_notmuch_new_query
    return '(%s)' % ' AND '.join('tag:%s' % tag for tag in get_notmuch_new_tags())
  File "/usr/lib/python3.6/site-packages/afew/NotmuchSettings.py", line 35, in get_notmuch_new_tags
    tags.remove("unread")
AttributeError: 'filter' object has no attribute 'remove'

```

```
❯ afew --version  
1.1.1.dev15+g7f942f3

❯ python --version                       
Python 3.6.2

❯ notmuch --version
notmuch 0.24.2

```

yang-ling:
Same error here.

1. afew version: 1.1.1.dev17+g090b511
2. Python version: 3.6.2
3. Notmuch version: 0.25

yang-ling:
I found that changing `tags = notmuch_settings.get_list('new', 'tags')` to `tags = list(notmuch_settings.get_list('new', 'tags'))` in `NotmuchSettings.py` can solve this problem.

GuillaumeSeren:
Hey,
I have encountered this error too while using python-3.x,
I think we have some issue with that version, so until those are fixed I suggest sticking with python-2.x.

But as I see there is already a PR, so if we can fix that it is good.

flokli:
fix using filters in #161.

PR #159, opened 2017-08-06
Fix AttributeError: 'filter' object has no attribute 'remove'

yang-ling:
Related issue is this one: https://github.com/afewmail/afew/issues/158

### DIFF
--- a/afew/NotmuchSettings.py
+++ b/afew/NotmuchSettings.py
@@ -30,7 +30,7 @@ def read_notmuch_settings(path = None):
     notmuch_settings.readfp(open(path))
 
 def get_notmuch_new_tags():
-    tags = notmuch_settings.get_list('new', 'tags')
+    tags = list(notmuch_settings.get_list('new', 'tags'))
     try:
         tags.remove("unread")
     except ValueError:


mkaito:
Why does notmuch_settings.get_list not return a list in the first place?

El 6 de agosto de 2017 7:23:05 GMT+01:00, SpicyCat <notifications@github.com> escribió:
>Related issue is this one: https://github.com/afewmail/afew/issues/158
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/afewmail/afew/pull/159
>
>-- Commit Summary --
>
>  * Fix AttributeError: 'filter' object has no attribute 'remove'
>
>-- File Changes --
>
>    M afew/NotmuchSettings.py (2)
>
>-- Patch Links --
>
>https://github.com/afewmail/afew/pull/159.patch
>https://github.com/afewmail/afew/pull/159.diff

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

yang-ling:
@mkaito I have no idea. I am not familiar with afew's source code. Today I updated afew and met this error. This PR is how I tried to fix it. Maybe it is not the best solution but it works for me.

mkaito:
The real question here is: why the heck does [this list comprehension](https://github.com/afewmail/afew/blob/b19a88fa1c06cc03ed6c636475cf4361b616d128/afew/configparser.py#L31) not return a list?

yang-ling:
@mkaito  I think [this line](https://github.com/afewmail/afew/blob/b19a88fa1c06cc03ed6c636475cf4361b616d128/afew/configparser.py#L36) causes this problem.

[In Python3, filter() returns return an object which is iterable, but not a list](https://stackoverflow.com/questions/12319025/filters-in-python3)

`filter(func,data) #python 2.x` is is equivalent to `list(filter(func,data)) #python 3.x`

flokli:
@mkaito In theory, I'm more for passing iterators around, so a filter would work here, and also allow us to get of the try/except. But I agree, the name of the `get_list` methid is really misleading and should be changed.

TBH, I'd like to see the whole custom config parsing code removed and something like ConfigObj with validators being used, even if this means the config format needs to be changed slightly.

yang-ling:
@flokli  Hi. At first thank you for creating afew.

At this moment, afew is broken in Python3 environment.  This PR is not elegant but it works for me.

Another solution is let [get_list](https://github.com/afewmail/afew/blob/b19a88fa1c06cc03ed6c636475cf4361b616d128/afew/configparser.py#L27) really returns list. But you said ` I'm more for passing iterators around` so I guess this solution won't be chosen.

So can you accept this PR? At least this PR can make afew working in Python3 environment.
Or do you have other plan?

GuillaumeSeren:
Hey,
If this fix at least momentarily the issue on python-3.x, it could be useful to have it to keep the tool working, for those around using that version of python.

As long as we don't break the support for python-2 as well.

GuillaumeSeren:
Ok I have tested the patch, on my local version (with python-2.x),
it worked so I guess we can use it to fix the problem, until we go for something better.

flokli:
closed in favor of #161 .

PR #163, opened 2017-09-17
Fix badge's links

Titan-C:
I clicked in the badges on the top of the readme and I was sent to the wrong page. The releases and travis links where switched. This fixes it


flokli:
right, thanks!

Issue #64, opened 2013-07-25
dependency on subprocess32


ff2000:
afew fails to run because it misses the package "subprocess32". As I understand [1] it's a port of the python-3.2 "subprocess" module to python 2.*. When I run with python3.2 this IMHO should not needed. Also the sources only import subprocess.
Could this requirement be made optional in setup.py? Or should that be simply patched out when installing for python3.2? I removed the subprocess32-line from "requires.txt" and at least it does not abort anymore with
pkg_resources.DistributionNotFound: subprocess32

[1] http://code.google.com/p/python-subprocess32/

lucc:
It seems that subprocess32 is only mentioned in the setup.py and not in the code any longer. And if I can run it with python2 just fine even without installing subprocess32. So I assume that can be removed from setup.py.

GuillaumeSeren:
hey @ff2000, ass @lucc reported it seems not to used in the code,
so to me it is ok to remove it from the 'setup.py' file,
if later we want to use it we will consider to add it.

PR #164, opened 2017-10-10
Remove subprocess32 dependency per issue 64

EricIO:
The subprocess32 package is no longer used in code so can therefore be removed.

Per https://github.com/afewmail/afew/issues/64


GuillaumeSeren:
Merged thank you ! 👍

PR #166, opened 2017-11-04
Add docs for rename flag added in #91

brighid:
Previously the documentation was silent on this flag, which is an important one
for the users who want it.

### DIFF
--- a/docs/source/move_mode.rst
+++ b/docs/source/move_mode.rst
@@ -10,6 +10,7 @@ Here is a full sample configuration for move mode:
 
     [MailMover]
     folders = INBOX Junk
+    rename = False
     max_age = 15
 
     # rules
@@ -77,6 +78,18 @@ in the configuration section. By providing
 
 afew will only check mails at most 15 days old.
 
+Rename
+------
+
+When afew moves mail files, if the the `rename` option in the configuration is
+set afew will rename those files. You should probably do this if you're using
+the `mbsync` program to interact with your IMAP server.
+
+.. code-block:: ini
+
+    rename = True
+
+
 Limitations
 -----------
 
@@ -100,4 +113,3 @@ moving it? ;)
 feature. If you tag a mail with two tags and there is a rule for each
 of them, both rules will apply.  Your mail will be copied into two
 destination folders, then removed from its original location.
-
--- a/docs/source/move_mode.rst
+++ b/docs/source/move_mode.rst
@@ -81,9 +81,16 @@ afew will only check mails at most 15 days old.
 Rename
 ------
 
-When afew moves mail files, if the the `rename` option in the configuration is
-set afew will rename those files. You should probably do this if you're using
-the `mbsync` program to interact with your IMAP server.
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hey,
I use MailMover a lot but I didn't use this maybe, it did not affect my use case (offlineimap),
but I think the option need better documentation, so I suggest you develop a bit more.

Please precise what rename change look like (short) at least.

brighid:
@GuillaumeSeren I don't understand "Please precise what rename change look like (short) at least", can you try that again? Pull request #91 already added this feature; I'm just adding a basic description of what the feature does.

GuillaumeSeren:
@brighid Yeah sorry it wasn't clear,
basically I was just asking to put an example in the doc since the feature was not documented,
and maybe not really used, just to be clear about what it does.

flokli:
@brighid 

What about the following description for 'Rename'?

```
With this option being set to true, afew will construct a new filename, composed of a made-up id and the flags part of the original filename instead of keeping the existing one.
This is needed when moving mails in combination with the imap syncing tool `mbsync`, which gets confused if multiple messages with the same UID exist in the same folder.
```

GuillaumeSeren:
@brighid @flokli 
I agree with the description, but I think it could be dangerous to activate this option on big mailboxes that did not use mbsync, so it should be something like a warning to avoid user bad configuration.
At least make it clear to not activate that if not using mbsync.

flokli:
I don't really see a problem here. Default is `false`, and a user needs to explicitly enable it.

Only mails that are moved will get renamed, and even with it being enabled, other tools like offlineimap shouldn't have problems with it either (they will see a mail being deleted in one folder, and appearing in another).

brighid:
Concurring with @floki: as a user, what I care about is that the option _not be undocumented._ Terse, manpage-style documentation is much better than no documentation. I'm not here to bikeshed the wording of it.

flokli:
Docs are better than no docs ;-)

TBH, I didn't know the feature even existed, and recently wished it existed ;-)

Can you update your PR with the suggested description?

GuillaumeSeren:
@brighid Well this is fine to me thank you for upgrading the patch 🍺

docs/source/move_mode.rst
```diff
@@ -81,9 +81,16 @@ afew will only check mails at most 15 days old.
 Rename
 ------
 
-When afew moves mail files, if the the `rename` option in the configuration is
-set afew will rename those files. You should probably do this if you're using
-the `mbsync` program to interact with your IMAP server.
+Set this option if you are using the `mbsync` IMAP syncing tool.
+`mbsync` adds a Unique IDentifier to files' names when it syncs them.
```
flokli: typo in `IDentifier`

brighid: I did that on purpose, trying to be clear about what "UID" meant. Not particularly attached to it, though — I was just trying to explain the feature without diving too deep into `mbsync` implementation details.

PR #169, opened 2017-11-26
Add filter tagging mail sent directly to me

aidecoe:
Add filter tagging mail sent directly to any of addresses defined in
Notmuch config file: `primary_email` or `other_email`.

Default tag is `to-me` and can be customized with `me_tag` option.


GuillaumeSeren:
Hey @aidecoe 🍺 
I've reviewed the patch and I was looking for something like that,
and it is a new filter so, no conflict with other features.

PR #170, opened 2017-11-27
Add filter verifing DKIM signature

aidecoe:
Filter tags e-mail having DKIM header with either `dkim-ok` or
`dkim-fail`.

### DIFF
--- a/afew/filters/DKIMValidityFilter.py
+++ b/afew/filters/DKIMValidityFilter.py
@@ -1,0 +1,39 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: ISC
+# Copyright (c) Amadeusz Zolnowski <aidecoe@aidecoe.name>
+
+from __future__ import print_function, absolute_import, unicode_literals
+
+import dkim
+
+from .BaseFilter import Filter
+
+
+def verify_dkim(path):
+    '''
+    Verify DKIM signature of an e-mail file.
+
+    :param path: Path to the e-mail file.
+    :returns: Whether DKIM signature is valid or not.
+    '''
+    with open(path, 'rb') as message_file:
+        message_bytes = message_file.read()
+
+    return dkim.verify(message_bytes)
+
+
+class DKIMValidityFilter(Filter):
+    '''
+    Verifies DKIM signature of an e-mail which has DKIM header.
+    '''
+    message = 'Verify DKIM signature'
+    header = 'DKIM-Signature'
+
+    def __init__(self, database, ok_tag='dkim-ok', fail_tag='dkim-fail'):
+        super(DKIMValidityFilter, self).__init__(database)
+        self.dkim_tag = {True: ok_tag, False: fail_tag}
+
+    def handle_message(self, message):
+        if message.get_header(self.header):
+            dkim_ok = all(map(verify_dkim, message.get_filenames()))
+            self.add_tags(message, self.dkim_tag[dkim_ok])
--- a/afew/tests/test_settings.py
+++ b/afew/tests/test_settings.py
@@ -27,6 +27,7 @@ class TestFilterRegistry(unittest.TestCase):
         from afew import FilterRegistry
         self.assertEqual(sorted(['FolderNameFilter',
                            'ArchiveSentMailsFilter',
+                           'DKIMValidityFilter',
                            'DMARCReportInspectionFilter',
                            'InboxFilter',
                            'SpamFilter',
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ def get_requires():
     if os.environ.get('TRAVIS') != 'true':
         yield 'notmuch'
     yield 'chardet'
+    yield 'dkimpy'
 
 setup(
     name='afew',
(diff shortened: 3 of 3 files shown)


flokli:
@aidecoe Thanks again for the contribution - will have a look ASAP :-)

Do you by any chance have the time to add a small testing harness, so we could run a chain of filters with a specific configuration on an email and check if tags get applied correctly?

I'd like to refactor some parts of the config parsing area (which is quite a mess), but am afraid this will break other things we might not catch…

afew/filters/DKIMValidityFilter.py
```diff
@@ -0,0 +1,57 @@
...
+    header = 'DKIM-Signature'
+
+    def __init__(self, database, ok_tag='dkim-ok', fail_tag='dkim-fail'):
+        super(DKIMValidityFilter, self).__init__(database)
+        self.ok_tag = ok_tag
+        self.fail_tag = fail_tag
+
+    def handle_message(self, message):
+        if message.get_header(self.header):
+            dkim_ok = all(map(verify_dkim, message.get_filenames()))
```
flokli: not sure about `message.get_filenames()`. This could yield multiple file names, and we need to open the file in python, which [can easily be done wrongly](https://git.notmuchmail.org/git?p=notmuch;a=blobdiff;f=bindings/python/notmuch/message.py;h=d5b98e4fdfdf5ab19e41fcda61ce0fc57d7b301f;hp=cce377d02a8024608bcf840252faae191d937acb;hb=91fe20cd90ce46bf80c416a5c4451f76b4d69ec5;hpb=073188e6905f1b5a065f696c4d02d0d1b3a7d769).

Is it possible to simply pass `message.get_message_parts()` to `dkim.verify`?

afew/filters/DKIMValidityFilter.py
```diff
@@ -0,0 +1,57 @@
...
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+# OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+#
```
flokli: Can you use [SPDX](https://wiki.spdx.org/view/Technical_Team/SPDX_Meta_Tags), and simply add a single
```
# SPDX-License-Identifier: ISC
```
instead of the license text?
I still have a PR lying around fixing this in other places as well :-)

flokli: @aidecoe see https://github.com/afewmail/afew/pull/171 for the PR.

afew/filters/DKIMValidityFilter.py
aidecoe: This is different. `dkim.verify` takes full raw e-mail message (as bytes, not as string). `get_filenames()` may yield multiple messages that's why for sanity `all()` is used, so we don't have false positive.

aidecoe:
@flokli It would be nice to have some test suites but I don't have too much spare time. Of course I will provide test suites for any other changes I will make in the future if the framework is in place.

afew/filters/DKIMValidityFilter.py
aidecoe: In other words that's the way to do it and there's no issue here.

flokli:
There are currently conflicts. Could you rebase?

In addition, I'd prefer to use `message.get_filename()`, which randomly picks one file location to open instead of mapping over all. In theory, the messages shouldn't differ (and if they do regardless, we'd have other problems :-) )

If you don't have the time to add tests, could you still add an example of a message with a valid DKIM signature, so we can add this to the test harness later?

aidecoe:
> In addition, I'd prefer to use message.get_filename(), which randomly picks one file location to open instead of mapping over all. In theory, the messages shouldn't differ (and if they do regardless, we'd have other problems :-) )

The filter verifies integrity of the message. So if any of those is wrong then that's integrity violation which should be signalled.

GuillaumeSeren:
Hey @aidecoe 
can you rebase the branch ?

aidecoe:
@GuillaumeSeren, I believe it is rebased onto afewmail master, isn't it?

GuillaumeSeren:
@aidecoe Ah ok I was a bit confused but the changes requested.

@flokli Completing the test suite is a long task and we don't have it right now.
I want to merge this PR and #175,to move forward and tag the 1.3.0,
we have small manpower let's not discourage users, and if we merge something bad we can fix it.

GuillaumeSeren:
👍 🍻

PR #171, opened 2017-11-28
Unify utf-8 comments, SPDX-License-Identifier, cleanup shebangs

flokli:
This uses a consistent utf-8 coding style comment, uses `SPDX-License-Identifier` instead of the full license quote (preserving copyright notices). and removes an unnecessary shebang in `afew.commands`.


GuillaumeSeren:
Hey @flokli 
I've seen you use that shortened licence on an other PR,
I think it is nice, because take less space in files.

I checked the diff looks ok,

We also should add a (global) licence file to the project,
because GitHub seems to no detect it well, and so not display it,
maybe also adding a small mention in the README that way it is clear for everyone.

flokli:
Added a `LICENSE` file, together with updated information in `setup.py`.
It might be `ISC` is not directly detected by github, but IMHO this is now already pretty clear.

andir:
LGTM.

PR #173, opened 2017-12-02
Inspect DMARC reports and tag the message

aidecoe:
DMARC reports usually come in ZIP files. To check the report you have to
unpack and search thru XML document which is very tedious. The filter
tags the message as follows:

- if there's any SPF failure in any attachment, tag the message with
"dmarc-spf-fail" tag, otherwise tag with "dmarc-spf-ok"
- if there's any DKIM failure in any attachment, tag the message with
"dmarc-dkim-fail" tag, otherwise tag with "dmarc-dkim-ok"


GuillaumeSeren:
Hey @aidecoe thank you for the patch
This looks good to me and it is a external filter, so let's merge it.

flokli:
@aidecoe Same as in #170: Could you please add an example mail for each of the 4 different tags, so they can be intergrated in the test harness?

andir:
Why I am the author of that commit? That looks wrong o.O

GuillaumeSeren:
Be careful @andir it's a trap we will ask you to come back on the project ! 🍺

GuillaumeSeren:
But I think you are just the first maintainer to validate the review so GH take the first one

aidecoe:
@andir has taken all the credit for my DMARC filter so now he provides example e-mail messages.

andir:
@aidecoe I removed myself from the `Author` field which I never added...

GuillaumeSeren:
@andir Yes it is automatically added when you review so, joke aside not a big deal,
but I hope this will not prevent you to review patches again in the future !

You can reply instead of approving if don't want your name in the commit 🍺 😄 🍻

PR #179, opened 2018-02-11
Support tags_blacklist in MeFilter

freeekanayaka:
It's not currently possible to use tags_blacklist in MeFilter,
although tags_blacklist is advertised as generic Filter setting, and
makes sense for MeFilter as well.


freeekanayaka:
Thanks @GuillaumeSeren

PR #183, opened 2018-03-04
Check for samefile error

### LINKED ISSUE #182: afew --move-mails fail with "SameFileError"
I'm using python 3.5 and afew 1.3.0 from debian, and afew try to move a mail onto itself.

Adding 

                    except shutil.SameFileError:
                        continue

at the correct location in MailMover.py (before the except shutil.Error as e) solve this problem.

vanicat:
Fix #182

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -73,6 +73,8 @@ class MailMover(Database):
                     try:
                         shutil.copy2(fname, self.get_new_name(fname, destination))
                         to_delete_fnames.append(fname)
+                    except shutil.SameFileError:
+                        continue
                     except shutil.Error as e:
                         # this is ugly, but shutil does not provide more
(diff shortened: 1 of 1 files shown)


flokli:
Can you add a log statement?

vanicat:
I'm not sure what you wish for. Is my commit message to terse?

flokli:
What I meant was adding a `logging.warning` in that case, so the user sees a message instead of just silently skipping over.

But of course, a more verbose commit msg would also be nice :-)

vanicat:
Ok, I've force-pushed a new commit that log such error.

Note that the "already exist" except do not warn the user of the problem.

GuillaumeSeren:
Hey @vanicat 
Thank you for the patch :beer:

PR #186, opened 2018-03-11
Check for shutil.SameFileError

### LINKED ISSUE #182: afew --move-mails fail with "SameFileError"

vanicat:
When copying a mail for MailMover, shutil.copy2 may failed because the
origin and source file are the same. In such a case, we should just
ignore the exception.

Fix #182


GuillaumeSeren:
Thank you @vanicat :beer:

Issue #139, opened 2017-04-25
Ability to skip notmuch hooks when running MailMover


jhelwig:
If `afew --move-mail` is added to the notmuch `pre-new` hook, you'll end up potentially fork-bombing yourself as the hook ends up being called over and over again due to MailMover running `notmuch new` (https://github.com/afewmail/afew/blob/master/afew/MailMover.py#L118). It would be nice if there was a configuration option for MailMover to tell it to pass `--no-hooks` to `notmuch new` to prevent the infinite recursion.

GuillaumeSeren:
Hey @jhelwig 
I use the mail-mover without the 'pre-new' hook so it is run in the end,
beside being a use-case / configuration problem, I think we should decide if you may be able to run mail-mover multiple time in a single call of afew, to me it's no but maybe there other's use case where it is usefull.

jhelwig:
For some context, what got me down the road of trying to call `afew --move-mail` from `pre-new` was the combination of "call afew --move-mails in an offlineimap presynchook" in the afew readme (https://github.com/afewmail/afew/blob/master/README.rst#move-mode) and notmuch's description of the `pre-new` hook ("Typically  this  hook  is used for fetching or delivering new mail to be imported into the database." https://notmuchmail.org/manpages/notmuch-hooks-5/).

GuillaumeSeren:
Hey,
I think the user is in control, but yeah running only once the 'notmuch new' should be the default,
so maybe if it was run, we should just prevent it to run again ?

If you could provide a patch it would be easier to solve I think.

PR #187, opened 2018-03-11
Add an --no-new option to afew --move-mails to not run notmuch-new

vanicat:
I had difficulties because `afew --move-mails` run `notmuch new` several time, and this could create difficulties if `notmuch new` has `afew --tag` in hooks or anything that change tags. So i tried to add a option to not run `notmuch new`

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -19,7 +19,7 @@ class MailMover(Database):
     '''
 
 
-    def __init__(self, max_age=0, rename = False, dry_run=False):
+    def __init__(self, max_age=0, rename = False, dry_run=False, no_new=False):
         super(MailMover, self).__init__()
         self.db = notmuch.Database(self.db_path)
         self.query = 'folder:{folder} AND {subquery}'
@@ -31,6 +31,7 @@ class MailMover(Database):
                                                        now=now.strftime('%s'))
         self.dry_run = dry_run
         self.rename = rename
+        self.run_new = not no_new
 
     def get_new_name(self, fname, destination):
         if self.rename:
@@ -106,7 +107,8 @@ class MailMover(Database):
         Update the database after mail files have been moved in the filesystem.
         '''
         try:
-            check_call(['notmuch', 'new'])
+            if self.run_new:
+                check_call(['notmuch', 'new'])
         except CalledProcessError as err:
             logging.error("Could not update notmuch database " \
                           "after syncing maildir '{}': {}".format(maildir, err))
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -75,6 +75,10 @@ options_group.add_argument(
     help="don't change the db [default: %(default)s]"
 )
 options_group.add_argument(
+    '-N', '--no-new', default=False, action='store_true',
+    help="do not run notmuch new after moving mails, you will have to run it manualy"
+)
+options_group.add_argument(
     '-R', '--reference-set-size', type=int, default=1000,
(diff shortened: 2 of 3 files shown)


GuillaumeSeren:
Hey @vanicat 
That seems useful to me, to have a way to control a bit better the process,
if we go this way maybe that would be nice to have an option to configure the default behaviour,
what do you think ?

vanicat:
Yes, it would be a good idea, but I don't know how to do it.

afew/commands.py
```diff
@@ -74,6 +74,10 @@
     '-d', '--dry-run', default=False, action='store_true',
     help="don't change the db [default: %(default)s]"
 )
+options_group.add_argument(
+    '-N', '--no-new', default=False, action='store_true',
+    help="do not run notmuch new after moving mails, you will have to run it manualy"
```
flokli: I'd suggest something more explaining like this:
> do not run `notmuch new` after moving mails, to avoid looping if afew is configured as a notmuch hook

I'd strip the `-N`, as it doesn't really speak for itself.

afew/MailMover.py
```diff
@@ -19,7 +19,7 @@ class MailMover(Database):
     '''
 
 
-    def __init__(self, max_age=0, rename = False, dry_run=False):
+    def __init__(self, max_age=0, rename = False, dry_run=False, no_new=False):
```
flokli: I'd prefer to call this variable `skip_notmuch_new`. It's a bit more explaining.

afew/MailMover.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, max_age=0, rename = False, dry_run=False):
                                                        now=now.strftime('%s'))
         self.dry_run = dry_run
         self.rename = rename
+        self.run_new = not no_new
```
flokli: assign to self.skip_notmuch_new, and don't invert logic here.

afew/MailMover.py
```diff
@@ -106,7 +107,8 @@ def __update_db(self, maildir):
         Update the database after mail files have been moved in the filesystem.
         '''
         try:
-            check_call(['notmuch', 'new'])
+            if self.run_new:
+                check_call(['notmuch', 'new'])
```
flokli: I'd prefer to do this check outside, everywhere where`__update_db` is called (similar to how dry_run is handled)

flokli:
Thanks for the PR! Let me first link #139, which would get fixed by this.

flokli:
In practice, we probably should still run `notmuch new` after moving mails around, but with the `--no-hooks` option. Any opinions?

GuillaumeSeren:
@flokli Hey,
> In practice, we probably should still run notmuch new after moving mails around, but with the --no-hooks option. Any opinions?
Yes, but now this option is called '--no-new' not '--no-hooks',
I have to reread the issue #139 because it was a long time ago :timer_clock: 

But I think we could also want to add a configuration preset of this behaviour,
the 'notmuch new' command is quite expensive and so it convenient to be able to choose when to run it.

flokli:
The issue we're trying to fight here is a cycle from `afew` to `notmuch new` to notmuch hooks to afew again, which could be interrupted by skipping hooks.

In practice, we don't want to have notmuch db in a wrong state, still assuming mail files are where they aren't anymore. So I'm against an option to disable putting the DB back into a sane state.

I digged a bit further, found an alternative solution, and documented it in #188.

vanicat:
Your proposition seem better, and should solve my problems.

Issue #188, opened 2018-03-21
Rework mail move logic


flokli:
Currently, when moving files, we move the file in the filesystem and shell out to `notmuch new`, and pick up new file locations in doing that.

This causes problems, as users often call `afew` via `notmuch`'s hook system, and in general, it's a bit invasive operation.

A better way would be to do the following on mail move:

 - Copy the file to the new location
 - Call `notmuch.database.add_message()` with the new location
 - Remove the file on the old location
 - Call `notmuch.database.remove_message()` with the old location

This would supersede #187, and fix #139.

GuillaumeSeren:
Hey @flokli ,
I think the logic is better than calling a big 'notmuch new', 
but it is a quite different from the #187 approach,
so the PR has to be changed a lot,
so I am checking if I can work a small patch to do it.

GuillaumeSeren:
I have started to work on the patch last night,
it went well but 'add_message()' is deprecated so I have to use index_file(),
and actually the hard part is the filename param is 
```
        :param filename: should be a path relative to the path of the
            open database (see :meth:`get_path`), or else should be an
            absolute filename with initial components that match the
            path of the database.
```
Which will probably a bit more work to get done, but I am on it.

see: https://git.notmuchmail.org/git?p=notmuch;a=blob;f=bindings/python/notmuch/database.py;h=342d665a224761a2557ea00233f7d99c7dbae321;hb=HEAD#l426

GuillaumeSeren:
Hey !
So I've finished a working patch, I tested it on my setup it work well,
if you can test and/or send feedback it would be great !

mjg:
Caveat: This patch ignores the "rename" setting! Had to clean up quite a few mbsync problems because of that.

Also, maybe it could be rebased and cut into smaller chunks that are easier to review?

GuillaumeSeren:
Hey,
@mjg thank you for testing this patch it need more reviews, can you develop a bit more your problem with the rename setting ?

I've just rebased the branch please respond on #189 to help improve this patch.

PR #189, opened 2018-03-25
Rework move logic

GuillaumeSeren:
Following the #188 issue, we needed to change move() function,
to be able to move, and upgrade the database a bit more lightly:
* Copy the original file to the destination
* Add the new file to the database
* Remove the original file
* Remove the original mail from db


afew/MailMover.py
```diff
@@ -50,7 +50,10 @@ def move(self, maildir, rules):
         '''
         # identify and move messages
         logging.info("checking mails in '{}'".format(maildir))
-        to_delete_fnames = []
+        # Open DB R/W if are not in dry-run
+        if not self.dry_run:
+            self.db = notmuch.Database(self.db_path,
+                                       mode=notmuch.Database.MODE.READ_WRITE)
```
flokli: Can we set database mode inside `__init__`, as we already know there if we want run in dry-run or not?
This would avoid opening the database a second time here…

afew/MailMover.py
```diff
@@ -72,31 +75,45 @@ def move(self, maildir, rules):
                     if self.dry_run:
                         continue
                     try:
-                        shutil.copy2(fname, self.get_new_name(fname, destination))
-                        to_delete_fnames.append(fname)
-                    except shutil.SameFileError:
-                        logging.warn("trying to move '{}' onto itself".format(fname))
-                        continue
+                        logging.info("Copy original mail: "+fname)
```
flokli: please use `.format` instead of concatenating strings: https://pyformat.info/

afew/MailMover.py
```diff
@@ -72,31 +75,45 @@ def move(self, maildir, rules):
...
+                        logging.info("Path of mail: "+os.path.dirname(fname))
+                        logging.info("Copy to: "+destination)
+                        dbpath = self.db.get_path()
+                        logging.info("Path of DB: "+dbpath)
+                        relpath=os.path.relpath(os.path.dirname(destination), dbpath)
```
flokli: typo - should be `realpath` and `os.path.realpath`.
Is realpath calculation even necessary?

afew/MailMover.py
```diff
@@ -72,31 +75,45 @@ def move(self, maildir, rules):
...
+                        filename = os.path.basename(fname)
+                        logging.info("filename: "+filename)
+                        logging.info("Relative path DB -> file: " + relpath)
+                        # 1 COPY
+                        shutil.copy2(fname, self.get_new_name(
+                                                            fname,
+                                                            destination))
+                        dest_file_rel = dbpath+'/'+relpath+'/'+filename
```
flokli: Please use `os.path.join`, or even better [pathlib](https://docs.python.org/3/library/pathlib.html) to join paths.

afew/MailMover.py
```diff
@@ -72,31 +75,45 @@ def move(self, maildir, rules):
...
+                        logging.info("Open DB R/W")
+                        # print vars(self.db)
+                        # 2 Add NEW MESSAGE TO DB
+                        moved_message = notmuch.Database.add_message(self.db, dest_file_rel)
+                        logging.info("Delete original mail file: "+fname)
+                        os.remove(fname)
+                        # 3 REMOVE ORIGINAL
+                        logging.info("Delete the original mail in DB")
+                        # 4 REMOVE ORIGINAL FROM DB
+                        notmuch.Database.remove_message(self.db, fname)
```
flokli: While I do favor some logging, this whole block seems to be wayy to extensive.

Most of the statements are merely printing out variable values. Maybe one or two lines with loglevel debug (something like `{} -> {}`.format(oldpath, newpath)`) should be enough

afew/MailMover.py
```diff
@@ -72,31 +75,45 @@ def move(self, maildir, rules):
...
+                    except IOError as e:
+                        logging.warn('Error IO: '+format(e))
                     except shutil.Error as e:
-                        # this is ugly, but shutil does not provide more
-                        # finely individuated errors
-                        if str(e).endswith("already exists"):
-                            continue
```
flokli: What's with this line? Do we already check somewhere else if source and destination are the same, which is why we skip that check?

afew/MailMover.py
```diff
@@ -59,7 +62,7 @@ def move(self, maildir, rules):
             messages = notmuch.Query(self.db, main_query).search_messages()
             for message in messages:
```
flokli: can we add below per-message code into a self-contained staticmethod? This would greatly improve readability and testability.

afew/MailMover.py
GuillaumeSeren: Yes I will move this to the __init__

afew/MailMover.py
GuillaumeSeren: ok I will upgrade those

afew/MailMover.py
GuillaumeSeren: In the beginning I was trying to use a relative path as explained in the doc, but after some time stuck with that I finally test and successfully made it work with full path.
So no this is not more necessary I will remove it

afew/MailMover.py
GuillaumeSeren: Ok I will upgrade that

afew/MailMover.py
GuillaumeSeren: Yes I agree but I will rework them to be more precise

afew/MailMover.py
GuillaumeSeren: I removed that because this will be logged by 'shutil' log and I think if you try to copy over the same thing it should be an error from shutil and/or your FS but I will add a test with a similar message for that case.

afew/MailMover.py
GuillaumeSeren: Not really sure to fully understand, but if you talk about export the whole per-message process I agree, I think the whole class need some refactor, for example the move() method is also searching the messages which is problematic to add tests and keep simple.

So I have project to refactor the class to be more simple, and support some unit tests, but as the mail mover is a big feature I think it would be safer to do that after.

afew/MailMover.py
flokli: Oh wow, there's a `os.path.relpath` as well - didn't know that.
But if the code now works with both relative and absolute paths, we shouldn't do this here, right.

GuillaumeSeren: Yes I discovered `relpath` aswell really useful to generate relative path.
I agree we can get rid of that now we only use absolute path.

afew/MailMover.py
GuillaumeSeren: Ah yes I agree the thing is I feel it would be simpler to use the function `move()` to do the actual moving operation, and create an other function like `getMessages()` to do the notmuch query and pass the message list to the move(), that way it could be easy tested by using mock data.

afew/MailMover.py
GuillaumeSeren: So about this point, I have tested several version, but if I open the database `READ_WRITE`,
in the `__init__` the call hang after 2 requests, and get stuck, I have to kill manually.
```
$ afew --move-mails -vvvvvvvv --all
INFO:root:checking mails in 'Perso/INBOX'                                                                                                                 
DEBUG:root:query: folder:Perso/INBOX AND tag:spam AND 576540000..1522686227                                                                               
DEBUG:root:query: folder:Perso/INBOX AND tag:sent AND 576540000..1522686227
```
So maybe it comes with my own version of notmuch / Xappian but unless I find a way to make it work, I think it is better to let the DB READ_ONLY all the time and just switch to READ_WRITE when needed (like I finally did), but it would nicer to have it setup only once.

GuillaumeSeren:
Hey @flokli I have upgraded the branch I have implemented all your feedback maybe beside the open connection ot the database in the `__init__` and also tested it with a lot of mails and it work fine.
If we aggree on the design I will rebase the branch to get the commit a bit cleaner.

flokli:
Already looks much cleaner, thanks!
Can you squash some of the follow-up commits together with the ones were I commented on? This makes reviewing a lot easier :+1: 
Additionally, in d3d62af you can use `%`-style format strings, and pass variables to the `logging.*` method directly, so we can strip some of the `.format` calls, too.

With that being done, I can have another look in why we need to open the database again for R/W.

afew/MailMover.py
flokli: Hanging while trying to open the database r/w a second time looks like the previous connection wasn't closed (since notmuch 0.23):

> When opening a database notmuch by default will wait for another process to release a write lock, rather than returning an error.

afew/MailMover.py
```diff
@@ -28,19 +29,23 @@ def __init__(self, max_age=0, rename=False, dry_run=False):
...
-                            str(uuid.uuid1()) + ':' + os.path.basename(fname).split(':')[-1]
-                        )
+                destination,
+                # construct a new filename,
+                # composed of a made-up ID
+                # and the flags part of the original filename.
+                str(
+                    uuid.uuid1())
+                    + ':'
+                    + os.path.basename(fname).split(':')[-1])
```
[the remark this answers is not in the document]
GuillaumeSeren: Yes sure, I just upgraded it.

GuillaumeSeren:
@flokli I am squashing the non moveMail() related commit.

Not really sure about the "you can use %-style format strings, and pass variables to the logging.* method directly, " but I have cleanup a bit of messages in the log.

GuillaumeSeren:
Hey @flokli 
If you do not have more feedback to fix in this patch I think we could push it,
sure there is no rush but there is a few other PR like #201 that should also appends,
ideally after this one, because it would force us to restart the review.

afew/MailMover.py
```diff
@@ -1,55 +1,58 @@
...
+    """ Move mails in folder maildir according to the given rules. """
     def move(self, maildir, rules):
         '''
         Move mails in folder maildir according to the given rules.
         '''
         # identify and move messages
         logging.info("checking mails in '{}'".format(maildir))
-        to_delete_fnames = []
         moved = False
         for query in rules.keys():
```
rpuntaie: Could you assure that the mailbox dirs do exist. I would like to only change the config, without having to additionally create a new directory manually:

```
    for ncs in 'tmp new cur'.split(): 
        destination = os.path.join(self.db_path,rules[query],ncs)
        os.makedirs(destination, mode=0o700, exist_ok=True)
```

afew/MailMover.py
```diff
@@ -58,44 +61,78 @@ def move(self, maildir, rules):
...
+                    else:
+                        moved = self.moveMail(fname, destination, True)
+        # Return the moved flag
+        return moved
+
+    """ Action of moving a mail and upgrade the DB """
+    def moveMail(self, mailFrom, mailTo, upgradeDatabase):
+        output = False
+        """ We need to check that we are not moving a file to itself """
+        if mailFrom == mailTo:
```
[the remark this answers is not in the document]
GuillaumeSeren: Yes sure I will rebase the branch and change that.

afew/MailMover.py
GuillaumeSeren: On that point I am not sure that behavior should be default,
to my pov, the mailmover should move from existing mailbox to an other (existing).

I am not sure, this case may append often, but if there is a real use case behind this,
we could easily add this check, feel free to add more details @rpuntaie maybe I am missing something.

GuillaumeSeren:
So, I have merged #200 and #201 this patch need some rebase / cleanup,
I will do it asap.

afew/MailMover.py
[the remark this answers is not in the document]
GuillaumeSeren: Hey,
after rebasing the branch, and re-reading I aggree we need to be sure that:
* mailFrom is a file AND mailTo is a directory

If this is not I will raise an error.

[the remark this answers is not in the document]
GuillaumeSeren: Did you see those:
```
            try:
                " mailFrom must be a file "
                if not os.path.isfile(mailFrom):
                    logging.error("mailFrom not a file: {}".format(mailFrom))
                    raise SystemExit
                " mailTo must be a directory "
                if not os.path.isdir(mailTo):
                    logging.error("mailTo not a dir: {}".format(mailTo))
                    raise SystemExit
```
So those tests already exist (sorry took me a second reading to be sure),
so duplicating those is not a good idea, and moving them end in the same state.

So tell me @rpuntaie

GuillaumeSeren:
Ok I have rebased the branch on the actual master,
I need help to test / review the patch.

Send reviews !

flokli:
We really need some sort of test suite in place ensuring behaviour doesn't break before merging this in, sorry!

GuillaumeSeren:
Hey @flokli ,
No need to be sorry, what we need is a project that can evolve and keep it's functionnal state,
yes having tests is better than not, no one is discussing that.

I will work on that and add a test around the MailMover, this issue is real and has to fixed.

afew/MailMover.py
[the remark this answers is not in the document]
rpuntaie: It seems like the commits I referred to have gone. There is not `mailFrom` anywhere anymore. So this seems to be obsolete.

As commented further below: tests are absolutely needed, to avoid break established behavior. E.g. I've seen `{}/{}/cur/` is now `{}/{}`, which is an interface change: The user would have to change its config.

PR #195, opened 2018-04-28
resolve #194 using shlex.split

### LINKED ISSUE #194: FolderNameFilter folder_explicit_list doesn't allow for spaces in folder names
When customizing the `FolderNameFilter` the option `folder_explicit_list` is parsed using `str.split()`. This prevents the inclusion of folders that have spaces, such as Gmail's "Sent Mail" folder. An easy fix would be to split using `shlex.split(str)` then one could enclose the folder name in quotes:

```
[FolderNameFilter]
folder_explicit_list = folder1 folder2 'Sent Mail'
```

jeffrutledge:
resolve #194


GuillaumeSeren:
:+1:

PR #199, opened 2018-05-19
doc: Fix filter config doc

arvindsv:
Filter config needs to have an index, not a name. Otherwise, it fails to match [this](https://github.com/afewmail/afew/blob/9a8309b30b536cffeeeb1901a3f0e33b11552ded/afew/Settings.py#L30) regex.


GuillaumeSeren:
Hey @arvindsv,
thank you for the patch !

Issue #43, opened 2012-11-13
catch NullPointerError in case mail file doesn't exist?


pazz:
```
[~] afew -t -a                                                              1 ↵  
Error opening /home/pazz/mail//home/pazz/Maildir/Sent/new/1345574901.M170094P32649Q1.brick:2,S: No such file or directory
Traceback (most recent call last):
  File "/home/pazz/.local/bin/afew", line 189, in <module>
    main(options, query_string)
  File "/home/pazz/.local/lib/python2.7/site-packages/afew/main.py", line 33, in main
    filter_.run(query_string)
  File "/home/pazz/.local/lib/python2.7/site-packages/afew/Filter.py", line 61, in run
    self.handle_message(message)
  File "/home/pazz/.local/lib/python2.7/site-packages/afew/filters/SpamFilter.py", line 32, in handle_message
    if message.get_header('X-Spam-Flag') == 'YES':
  File "/home/pazz/.local/lib/python2.7/site-packages/notmuch/message.py", line 233, in get_header
    raise NullPointerError()
notmuch.errors.NullPointerError

```

frankqux:
I too am getting this using `notmuch 0.19+103~g294bb6d`
```python
Traceback (most recent call last):
  File "/home/frankqux/.local/bin/afew", line 9, in <module>
    load_entry_point('afew==0.0.0', 'console_scripts', 'afew')()
  File "/home/frankqux/.local/lib/python2.7/site-packages/afew-0.0.0-py2.7.egg/afew/commands.py", line 205, in main
    inner_main(options, database, query_string)
  File "/home/frankqux/.local/lib/python2.7/site-packages/afew-0.0.0-py2.7.egg/afew/main.py", line 38, in main
    filter_.run(query_string)
  File "/home/frankqux/.local/lib/python2.7/site-packages/afew-0.0.0-py2.7.egg/afew/filters/BaseFilter.py", line 73, in run
    self.handle_message(message)
  File "/home/frankqux/.local/lib/python2.7/site-packages/afew-0.0.0-py2.7.egg/afew/filters/HeaderMatchingFilter.py", line 41, in handle_message
    value = message.get_header(self.header)
  File "/usr/local/lib/python2.7/dist-packages/notmuch/message.py", line 233, in get_header
    raise NullPointerError()
notmuch.errors.NullPointerError
```
Also, is this not the same as ~~#35~~ ?

varac:
[...]

PR #200, opened 2018-06-12
allow to specify args for notmuch new

mjg:
In move mode, afew calls `notmuch new` after moving mails around. This
prevents `afew -m` from being used in a pre-new hook in `notmuch`.

Allow to specify args, so that `afew -m --notmuch-args=--no-hooks` can
live happily in a pre-new hook.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -34,11 +34,16 @@ class MailMover(Database):
 
     def get_new_name(self, fname, destination):
         if self.rename:
+            parts = os.path.basename(fname).split(':')
+            if len(parts) > 1:
+                flagpart = ':' + parts[-1]
+            else:
+                flagpart = ''
             return os.path.join(
                             destination,
                             # construct a new filename, composed of a made-up ID and the flags part
                             # of the original filename.
-                            str(uuid.uuid1()) + ':' + os.path.basename(fname).split(':')[-1]
+                            str(uuid.uuid1()) + flagpart
                         )
         else:
             return destination
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -34,6 +34,7 @@ class MailMover(Database):
 
     def get_new_name(self, fname, destination):
         basename = os.path.basename(fname)
+        submaildir =  os.path.split(os.path.split(fname)[0])[1]
         if self.rename:
             parts = basename.split(':')
             if len(parts) > 1:
@@ -43,7 +44,7 @@ class MailMover(Database):
                             # construct a new filename, composed of a made-up ID and the flags part
                             # of the original filename.
             basename = str(uuid.uuid1()) + flagpart
-        return os.path.join(destination, basename)
+        return os.path.join(destination, submaildir, basename)
 
     def move(self, maildir, rules):
         '''
@@ -54,7 +55,7 @@ class MailMover(Database):
         to_delete_fnames = []
         moved = False
         for query in rules.keys():
-            destination = '{}/{}/cur/'.format(self.db_path, rules[query])
+            destination = '{}/{}/'.format(self.db_path, rules[query])
             main_query = self.query.format(
                 folder=maildir.replace("\"", "\\\""), subquery=query)
             logging.debug("query: {}".format(main_query))
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -33,20 +33,17 @@ class MailMover(Database):
         self.rename = rename
 
     def get_new_name(self, fname, destination):
+        basename = os.path.basename(fname)
         if self.rename:
-            parts = os.path.basename(fname).split(':')
+            parts = basename.split(':')
             if len(parts) > 1:
                 flagpart = ':' + parts[-1]
             else:
                 flagpart = ''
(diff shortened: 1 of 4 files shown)


flokli:
Can you use #187 for that?

mjg:
> Can you use #187 for that?

First of all, sorry for not searching the issues before... #187 is not quite it but #188 looks promising since it updates the notmuch db (even more effectively) but avoids hooks. It implies a change in behaviour for those users which use hooks (in my case, welcome change). I'll try #188 and report back.

mjg:
So, that approach (use notmuch bindings, not the binary) works and looks better all together, but the implementation is lacking (ignores rename setting). Cleaning up that patch over there seems to be worthwhile if you decide that the change in behaviour is OK (no more hooks being fired - fine for me :)).

GuillaumeSeren:
Hey,
first thank you for the patch :beer: 

I agree on the feature as a heavy user of the MailMover myself,
it is close / same as the #187 and could trigger the same db non atomic operation,
but if we assume that a user can manage that issue on its own,
we could merge that before the rework of the MailMover.

I will check the patch, but i think if it can solve both your issue (#200) and (#187),
it could be a good thing.

@vanicat Can you have a look and check if it solve your issue ?

rpuntaie:
#200 is for `notmuch` `pre-new` hook. I call `afew` in `post-new` hook. There it does not lead to a second `afew`  run. It seems like `notmuch` sees its DB in use and so does not call `post-new`. 

I consider using it in the `post-new` hook is the right way,
- because `notmuch` needs to have its DB up to date before `afew` can work on it, else #43.
- and after `afew` moved mails, it cares to call `notmuch new` again, since again its DB must be updated to reflect `afew's` changes.

mjg:
> #200 is for `notmuch` `pre-new` hook. I call `afew` in `post-new` hook. There it does not lead to a second `afew` run. It seems like `notmuch` sees its DB in use and so does not call `post-new`.
> 
> I consider using it in the `post-new` hook is the right way,
> 
> * because `notmuch` needs to have its DB up to date before `afew` can work on it, else #43.
> * and after `afew` moved mails, it cares to call `notmuch new` again, since again its DB must be updated to reflect `afew's` changes.

This depends on your use of MUA and synchronisers. For me, the MUA only changes tags in the notmuch db or adds messages (sending), i.e. it operates through the nm library and the db is in sync *locally*.

afew/notmuch does the *remote* sync, which encompasses fetching/pushing of e-mail including necessary moves by tag and tagging by folder. This has to be done in a specific order to make sense. My ordering ist:

1. nm.pre-new: `afew -m --notmuch-args=--no-hooks` makes sure any new tags lead to apropriate foldering
2. nm.pre-new: `mbsync` pushes the local state to the remote, fetches remote updates
3. notmuch syncs the local file system state (changed by mbsync) to the db
4. nm.post-new: `afew -t -n` tags new messages in the db

I could not come up with any other ordering that makes sure that db and local state and remote state are in sync and that requires only one run of notmuch, mbsync and the two afew modules. But I'd be happy to learn alternatives.

rpuntaie:
I'm currently moving from Thunderbird to a more open mail store.
So I'm just figuring things out step-by-step and happy to learn from others.

I have set up a mail sync script, that basically does

1. mbsync
2. notmuch new

   in `$MAILDIR/.notmuch/post-new` I have

       # create missing folders based on afew config, then
       afew -mn
       afew -tn

My `[MailMover]` does not use tags, but rather something like:

    gmail/INBOX = 'from:someone AND subject:something':somefolder

`-tn` then tags based on the new folders using FolderNameFilter.

remote is synched only the next time my sync script is called.
It will then

- update the local storage with new mails from remote
- update remote with the mails I have removed locally (i.e. moved to non-synced local mailboxes)

I mean, remote and local will get out of sync anyway.
There is no need to leave the script with both at sync.

GuillaumeSeren:
Ok let's merge this thank you for the patch @mjg :coffee:

afew/MailMover.py
```diff
@@ -52,7 +56,7 @@ def move(self, maildir, rules):
         to_delete_fnames = []
         moved = False
         for query in rules.keys():
-            destination = '{}/{}/cur/'.format(self.db_path, rules[query])
```
flokli: Where did that `cur` here go?

mjg: Apparantly, #200 and #201 got merged together when they were both merged (or I rebased one PR branch too early, or github computes the diff relative to the original merge base).
In any case, this is reflected in the change to new_name(): In short, afew used to move mail from new to cur but failed to update flags (everyone moving to cur should update flags), so #201 made afew leave mail in the respective subdir rather than chopping off new/cur/.. and appending cur..

flokli: Sorry, but this still doesn't feel quite readable to me.

I could guess `submaildir` is `cur` or `new`, but just taking `[1]` out of `os.path.split()`'s `tail` sounds scary - what happens to the other path components?

Could this maybe be made slightly more readable way, maybe by adding some comments?

We could also probably use `pathlib` paths around instead of strings.

Issue #190, opened 2018-04-09
overwrapping path in maildir


tanyo13:
Hello,

Move-mail function can abort when mailboxes have overwrapping path names.

I have entries something like the following:

[MailMover]
folders = main/INBOX main/INBOX.archive
main/INBOX = 'NOT tag:inbox':main/INBOX.archive
main/INBOX.archive = 'tag:deleted':main/INBOX.Trash

When two files for a message that matches 'NOT tag:inbox' exist, one in main/INBOX and the other in main/INBOX.archive, afew --move-mail fails with the following stack trace:

  File "/usr/local/bin/afew", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/site-packages/afew/commands.py", line 159, in main
    inner_main(args, database, query_string)
  File "/usr/local/lib/python3.6/site-packages/afew/main.py", line 33, in main
    mover.move(maildir, rules)
  File "/usr/local/lib/python3.6/site-packages/afew/MailMover.py", line 75, in move
    shutil.copy2(fname, self.get_new_name(fname, destination))
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/shutil.py", line 257, in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/shutil.py", line 104, in copyfile
    raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
shutil.SameFileError: '.../main/INBOX.archive/cur/1523257381_0.1544....' and '.../main/INBOX.archive/cur/1523257381_0.1544....' are the same file

This is I think because selecting the file in the specified maildir is done using "if maildir in name".  Around line 63 in method move() in MailMover.py, to_move_fnames is defined as:
      to_move_fnames = [name for name in all_message_fnames if maildir in name]
If you replace "maildir in name" is with "maildir + '/cur' in name", the problem is gone, though I do not know if that is a correct fix.

mjg:
The correct fix should check that both paths have the same leading components up to maildir:
```
fullmaildirs = [ os.path.join(self.db_path, maildir, subdir) for subdir in ['new', 'cur'] ]
any([ fullmaildir  == os.path.commonprefix(fullmaildir, name) for fullmaildir in fullmaildirs ])
```
Or a regex...

PR #201, opened 2018-06-14
Mail mover fix renames

### LINKED ISSUE #196: Should afew's MailMover move emails out of Maildir's /new or /tmp subfolders ?
I use afew as a post-new notmuch hook, after I've synced my mail using [mbsync](http://isync.sourceforge.net/mbsync.html) but before my new mail has been read by my MUA. As a result, new mail is still in the /new sub-directory (and indexed as such by notmuch). 

This *also* means that each of those mail's filenames has not yet had any metadata appended, which means that filename doesn't have a colon in it, which means that MailMover's [file renaming strategy](https://github.com/afewmail/afew/blob/9a8309b30b536cffeeeb1901a3f0e33b11552ded/afew/MailMover.py#L41) breaks.  The end result is that the UID element of the mail is not stripped out, leading to conflicting UIDs in the destination Maildir.

I have a small patch for this for afew, which limits moves to files in the "/cur" subfolder. 

Does that seem the right approach? Or is this something better handled by calling afew at a different point?

mjg:
So here is a suggestion to fix #196.
A fix to #190 could go on top easily (although it is unrelated). In fact, there is currently a mix of "treating paths right" (`os.path`...) and working with `/` by hand which could use some sanitizing.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -34,11 +34,16 @@ class MailMover(Database):
 
     def get_new_name(self, fname, destination):
         if self.rename:
+            parts = os.path.basename(fname).split(':')
+            if len(parts) > 1:
+                flagpart = ':' + parts[-1]
+            else:
+                flagpart = ''
             return os.path.join(
                             destination,
                             # construct a new filename, composed of a made-up ID and the flags part
                             # of the original filename.
-                            str(uuid.uuid1()) + ':' + os.path.basename(fname).split(':')[-1]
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hey @mjg 
Thank you for the PR, as the MailMover is actually in refactor and the PR is open #189,
we have to state the order to pass the PR but we are working on it.

GuillaumeSeren:
As the mail mover as to be refactor and discussed again, 
I think we can merge this.

Issue #68, opened 2014-02-18
Where is register_filter?


larsks:
`docs/source/extending.rst` suggests that custom filters in `~/.config/afew` should start with:

```
from afew.Filter import Filter, register_filter
```

But there is no `afew.Filter` module nor is there a `register_filter` function defined anywhere.

matclab:
Same problem with version 1.3.0 here:
```
INFO:root:Importing user filter 'snooze.py'
Traceback (most recent call last):
  File "/usr/sbin/afew", line 11, in <module>
    load_entry_point('afew==1.3.0', 'console_scripts', 'afew')()
  File "/usr/lib/python3.6/site-packages/afew/commands.py", line 135, in main
    __import__(file_name[:-3], level=0)
  File "/home/clabaut/.config/afew/snooze.py", line 1, in <module>
    from afew.Filter import Filter, register_filter
ModuleNotFoundError: No module named 'afew.Filter'
```

matclab:
Doing:
```
from afew.filters.BaseFilter  import Filter                                     
from afew.FilterRegistry import register_filter 
```
solves the problem.

@flokli Should the doc be updated ? Or is there a missing export of those symbols
 ?

flokli:
@matclab imports are currently a bit funky at least, requiring a lot of changes.

I'd say, updating docs to the current state for now would be the most sensible thing to do. Can you file a PR?

matclab:
Yes, I'll try to do it in the weekend.

Le sam. 23 juin 2018 à 14:56, Florian Klink <notifications@github.com> a
écrit :

PR #202, opened 2018-06-25
Update extension example with working import

matclab:
Provide a fix for #68 (was already closed, but there was a similar problem).


flokli:
thanks!

PR #203, opened 2018-07-21
Use `db.find_message` in `Filter.commit`

samdoshi:
Use `notmuch.Database.find_message` instead of `notmuch.Query` as a more idiomatic way to get a message and to avoid problems with escaping double quotes.

---

A little background... I tried using `afew` a few months back, but ran into crashing bugs with `InboxFilter`, I've spent some time this morning trying to get to the bottom of them.

Basically some of my messages have `id` tags with a `"` in them. I tried escaping the double quote, but there isn't that much info on it, I think you're supposed to repeat them (e.g. `"` becomes `""`). However using [`Database.find_message`][find_message] is a better solution, and may even have better performance.

[find_message]: https://notmuch.readthedocs.io/projects/notmuch-python/en/latest/database.html#notmuch.Database.find_message

### DIFF
--- a/afew/filters/BaseFilter.py
+++ b/afew/filters/BaseFilter.py
@@ -103,16 +103,15 @@ class Filter(object):
             db = self.database.open(rw=True)
 
             for message_id in dirty_messages:
-                messages = notmuch.Query(db, 'id:"%s"' % message_id).search_messages()
+                message = db.find_message(message_id)
 
-                for message in messages:
-                    if message_id in self._flush_tags:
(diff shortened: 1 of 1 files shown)


flokli:
Thanks!

samdoshi:
Thanks for the quick response!

Any idea (roughly) when the next release might be tagged?

flokli:
I can do a release later today or (more likely tomorrow).

Let's wait for some quick feedback on the related issue :-)

Issue #181, opened 2018-02-20
Verbose mode isn't verbose


varac:
```
  -v, --verbose         be more verbose, can be given multiple times
```

I tried giving `-v` multiple times in order to get *any* verbosity, but it just don't change anything.

```
» afew -t lastmod:3480603..3480605 
» afew -vvvvvv -t lastmod:3480603..3480605
» 
```

were messages tagged or not ?

GuillaumeSeren:
Hey @varac 
To me this is working fine on master see #180, so please test against master and reopen if issue is still there.

varac:
Still not giving any output at all to me, running latest master. 

```
⏻  ~  » afew -vvvvvvvv -t --new
⏻  ~  »
```

Can anybody else please try to reproduce ? I cannot reopen this issue.

GuillaumeSeren:
Hey @varac please test the issue against the 1.3.0 ?

flokli:
@varac I still can't reproduce logs verbose mode not being verbose, sorry.
Note having 2 `-v` parameters is already enough for loglevel `DEBUG` ;-)

The posted traceback is helpful.
`search_messages()` from `notmuch.query` raised a `XapianError`, most likely due to a wrongly formatted query string.
You might want to add some more debug log statements inside `afew/main.py`, to get an idea where it's crashing, while checking your afew configuration for invalid queries.

flokli:
Not sure about the setup, but I'd suspect some old/broken versions lingering around somewhere and being loaded. As we can't reproduce, I'd close this one.

varac:
@floki, the bug is triggered when using python2 instead of python3, see https://github.com/afewmail/afew/issues/180#issuecomment-402825613.

Issue #180, opened 2018-02-13
dry-run not working


varac:
At least in combination with `--move-mails`, `--dry-run` is not working as expected:

```
» afew --move-mails -vvvvvvvv --dry-run                                                                                                             
» afew --move-mails -vvvvvvvv          
Processed 2 total files in almost no time.
No new mail. Detected 2 file renames.

» afew --version
1.3.0
```

GuillaumeSeren:
Hey @varac thank you for the feedback,
Yes I think it is a real issue, we should fix it asap.

varac:
@GuillaumeSeren Please fix this, it's important for debugging, right now I'm really stabbing in the dark with no dry-run and no way to run in verbose mode (#181).

GuillaumeSeren:
Hey @varac 
I understand your feedback and be sure that it is in my todo list,
but there is a lot of thing in there (+ everyday work).

So if you want this fast(-er) I suggest that you go ahead and do it on your own,
it would be helpful, and we could help / review when needed.

GuillaumeSeren:
Hey @varac 
I have investigate a bit more about your problem, and on my setup the exact command you report 'afew --move-mails -vvvvvvvv --dry-run' is verbose and not move anything, here an example of the output
```sh
$ afew --version                                                                                                                                                                                                 1.3.0
$ afew --move-mails -vvvvvvvv --dry-run
INFO:root:    b'[2017-11-27 16:46:06] Jon Mason | Re: [GIT PULL] NTB bug fixes for v4.15'                                                                                                                                                     
INFO:root:from 'Perso/archive' to 'Perso/INBOX'                                                                                                                                                                                               
INFO:root:I would move mail                                                                                                                                                                                                                   
INFO:root:    b'[2017-11-27 16:41:24] Maxime Ripard | [PATCH v2 00/18] drm/sun4i: Add A83t LVDS support'                                                                                                                                      
INFO:root:from 'Perso/archive' to 'Perso/INBOX'                                                                                                                                                                                               
INFO:root:I would move mail                                                                                                                                                                                                                   
INFO:root:    b'[2017-11-27 16:32:02] Lorenzo Pieralisi | Re: [PATCH] PCI: designware-ep: Fix ->get_msi() to check MSI_EN bit'                                                                                                                
INFO:root:from 'Perso/archive' to 'Perso/INBOX'                                                                                                                                                                                               
INFO:root:I would move mail
```

[...]

PR #204, opened 2018-08-05
Drop Python 2 support from docs, remove Python 2 hacks

flokli:
We already encourage running afew with Python 3 for some time now, and there's some weird bugs when running with Python 2 (#180, #181).

Given Python 3 is available almost everywhere too, let's get rid of python 2 and hacks present in the codebase.

cc @varac

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 dist: trusty
 language: python
 python:
-  - "2.7"
   - "3.4"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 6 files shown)


GuillaumeSeren:
Hey @flokli 
Yes sure that will cleanup some workaround for python2 but maybe also disturb some user was happily using python2 without issues, forcing them to upgrade.

But IMO it is ok to do it, but we will need to tag a major version to reflect the change (v2).

flokli:
Travis [doesn't seem to support python 3.7 (yet)](https://travis-ci.org/flokli/afew/jobs/480110607), would have added that too :-)

flokli:
@polyzen thanks for the pointer, follow-up in https://github.com/afewmail/afew/pull/217 .

Issue #97, opened 2014-12-27
MailMover is unable to handle spaces in folder names


seschwar:
Due to how afew's configuration is parsed the MailMover is unable to handle folder names containing spaces.

This is a problem for Gmail accounts synced with offlineimap which include folder name like, "[Google Mail]", "Sent Mail" and "All Mail".

Replacing the spaces with non-breaking spaces doesn't help since Python's `string.split()` and `re.split("\s+", ...)` treat them the same as regular spaces.

PR #206, opened 2018-09-19
Split folders and preserve quoted names

nicolasbock:
If a folder in the `MailMover` section includes a
space (e.g. `[Gmail].Sent Mail`) there is currently not way to quote
the folder name such as to preserver the space in it. This change
replaces the standard `split()` function with one that understands
quotes. For the above example, the following config entry works:

        folders = "[Gmail].Sent Mail" other_folder

Fixes: #97

### DIFF
--- a/afew/Settings.py
+++ b/afew/Settings.py
@@ -7,6 +7,7 @@ from __future__ import print_function, absolute_import, unicode_literals
 import os
 import re
 import collections
+import shlex
 
 from .configparser import SafeConfigParser
 from afew.FilterRegistry import all_filters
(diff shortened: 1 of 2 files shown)


flokli:
Thanks for your PR!

The config parser is something that I'd eventually like to see replaced with ConfigObj anyhow, but that will most likely introduce more format changes down the road anyhow ;-)

Can you also update `docs/source/move_mode.rst`, explaining it's possible to use quotes, with the example above?

nicolasbock:
Done @flokli .

flokli:
Thanks!

PR #216, opened 2019-01-08
Quote folder names in search query

nicolasbock:
When using folder names containing whitespace characters (see PR #206)
the folder part of the notmuch search query needs to be quoted.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -22,7 +22,7 @@ class MailMover(Database):
     def __init__(self, max_age=0, rename = False, dry_run=False):
         super(MailMover, self).__init__()
         self.db = notmuch.Database(self.db_path)
-        self.query = 'folder:{folder} AND {subquery}'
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hello,
thank you for the patch :clap:

flokli:
@nicolasbock I guess this breaks when the output contains double quotes - we just enclose quotes, but not escape.

Instead of this patch, can we escape [while setting `main_query` by calling `.format`](https://github.com/afewmail/afew/pull/216/files#diff-8651d0b5e8a0fa3360ac7effba638683L56) using [`shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote)? This should enclose single-quotes automatically:

```py
>>> import shlex
>>> print(shlex.quote("foo"))
foo
>>> print(shlex.quote("foo\""))
'foo"'
```

nicolasbock:
@flokli Something like #218 ?

PR #218, opened 2019-01-17
Shell quote mail folder

nicolasbock:
Follow on to PR #216:

In cases where the folder name contains characters that should be
quoted, use the `shlex.quote()` function to properly format the folder
name.

### DIFF
--- a/afew/MailMover.py
+++ b/afew/MailMover.py
@@ -5,6 +5,7 @@
 import notmuch
 import logging
 import os, shutil
+import shlex
 from subprocess import check_call, CalledProcessError
 
 from .Database import Database
@@ -22,7 +23,7 @@ class MailMover(Database):
     def __init__(self, max_age=0, rename = False, dry_run=False):
         super(MailMover, self).__init__()
         self.db = notmuch.Database(self.db_path)
-        self.query = 'folder:"{folder}" AND {subquery}'
+        self.query = 'folder:{folder} AND {subquery}'
         if max_age:
             days = timedelta(int(max_age))
             start = date.today() - days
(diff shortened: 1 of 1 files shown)


flokli:
@nicolasbock 
```
self.query = 'folder:"{folder}" AND {subquery}'
```
need to be un-quoted again, as `shlex.quote` does add enclosing quotes automatically when necessary:

```python
>>> import shlex
>>> print(shlex.quote("foo"))
foo
>>> print(shlex.quote("foo\""))
'foo"'
```

nicolasbock:
You are right, I missed that. It's fixed now.

flokli:
Thanks!

nicolasbock:
@flokli For some reason this is not working. With this PR the search query does not return anything.

Running `afew --move-mail --verbose --verbose` fails to find matching emails:
```
DEBUG:root:query: folder:'xxx.gmail/[Gmail].Spam' AND NOT tag:spam
[NO RESULTS]
```

But running that query in the shell works:
```
$ notmuch search folder:'xxx.gmail/[Gmail].Spam' AND NOT tag:spam
[RESULTS]
```

Interestingly, when I revert the `shlex.quote()` call in line [58](https://github.com/afewmail/afew/blob/master/afew/MailMover.py#L58) but change the query string in line [26](https://github.com/afewmail/afew/blob/master/afew/MailMover.py#L26) to
```python
self.query = "folder:'{folder}' AND {subquery}"
```

I get the same result as with this PR, _i.e._ the query does not return anything. The failure is reproducible with single quotes but I don't understand why.

Do you have any suggestions?

PR #219, opened 2019-01-20
add MAILDIR as fallback for database location

paretje:
Notmuch cli uses the MAILDIR environment variable as a default when no database path is configured in notmuch config. This adds the same behaviour to afew.

```
       database.path
              The top-level directory where your mail currently exists and to where mail will be delivered in the future. Files
              should be individual email messages. Notmuch will store its database within a sub-directory of the  path  config‐
              ured here named .notmuch.

              Default: $MAILDIR variable if set, otherwise $HOME/mail.
```

### DIFF
--- a/afew/Database.py
+++ b/afew/Database.py
@@ -20,7 +20,7 @@ class Database(object):
     def __init__(self):
         self.db_path = notmuch_settings.get('database', 'path',
                                             fallback=os.environ.get('MAILDIR',
-                                                                    '%s/mail' % os.environ.get('HOME')))
+                                                                    '{}/mail'.format(os.environ.get('HOME'))))
         self.handle = None
 
     def __enter__(self):
--- a/afew/Database.py
+++ b/afew/Database.py
@@ -7,7 +7,6 @@ from __future__ import print_function, absolute_import, unicode_literals
 import os
 import time
 import logging
-import configparser
 
 import notmuch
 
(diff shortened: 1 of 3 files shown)


afew/Database.py
```diff
@@ -17,7 +18,9 @@ class Database(object):
     '''
 
     def __init__(self):
-        self.db_path = notmuch_settings.get('database', 'path')
+        self.db_path = notmuch_settings.get('database', 'path',
+                                            fallback=os.environ.get('MAILDIR',
+                                                                    '%s/mail' % os.environ.get('HOME')))
```
flokli: can you use [PyFormat](https://pyformat.info/) here?

flokli:
I'm wondering, can we avoid reading notmuch settings on our own completely?

paretje:
I checked, and `notmuch_config_open` isn't exposed in libnotmuch. So unless we would use something like `notmuch config get database.path`, I don't think we can get rid of this. E.g. alot also reads the config file on its own.

flokli:
@paretje then fine for me - could you rebase this on master?

In addition, it would be nice if you could add a line to `docs/configuration.rst` mentioning that afew honors the `MAILDIR` environment variable as fallback, like notmuch does too.

paretje:
@flokli I just documented how afew determines the notmuch database location.

flokli:
Thanks!

Issue #221, opened 2019-02-18
tags = -unread does not work


agrudzinsky:
Any ideas why I cannot remove tag `unread` from my emails using afew?

I'm using afew version 1.3.0, notmuch version 0.28

Filter example:

```
[Filter.1]
query = subject:'testing afew'
tags = -unread
message = Testing afew
```

Another example
```
[Filter.1]
query = from:spammer@spam.com
tags = +spam
message = Spam

[Filter.2]
query = tag:spam
tags = -unread
message = Mark as read

[Filter.3]
query = tag:spam
tags = -new
message = Bypass inbox
```
The email is marked as `spam` and does not make it to `inbox`, but when I open unread messages, it's still there in my face.

Note, running `notmuch tag -unread tag:spam` from the command line works properly.

Is this a bug or a feature?

Thanks.

GuillaumeSeren:
Hello,
I think it is a feature to keep tag unread/read untouched by afew like you can see in the in the 'afew/filters/BaseFilter.py', I guess the idea is to keep that part for your MUA.
```python
    def remove_tags(self, message, *tags):
        if tags:
            filtered_tags = list(tags)
            try:
                filtered_tags.remove('unread')
                filtered_tags.remove('read')
            except ValueError:
                pass
            self.log.debug('Removing tags %s from id:%s' % (', '.join(filtered_tags),
                                                           message.get_message_id()))
            self._remove_tags[message.get_message_id()].update(filtered_tags)

```
You can easily add a spam tag or something like that if you want, like:
```
[Filter.10]
message = Twitter
query = from:'info@twitter.com'
tags = +twitter;+spam
```

Guillaume

agrudzinsky:
Guillaume, thanks for your reply.

Would you consider removing this limitation?  I do not see a good reason for it.  For notmuch, afew, and alot, `unread` is nothing more than a regular tag.  If notmuch allows to manipulate `unread`, I don't see why afew shouldn't.  I'd leave it to the users to decide whether they want to change `unread` automatically or explicitly.

Of course, I can work around this limitation using other tags or use a notmuch hook to remove `unread` tag from certain emails after running afew.  But why jump through the hoops?  

[...]

PR #231, opened 2019-03-20
Filter: let user remove "unread" or "read" tags

hydrargyrum:
If the users wants to do it with notmuch-tag, they can, so why
not with afew?

Closes https://github.com/afewmail/afew/issues/221


flokli:
@hydrargyrum can you update NEWS.md, adding a line with something like this

> Support for removing `unread` and `read` tags in filters

hydrargyrum:
@flokli done

flokli:
Thanks!

PR #233, opened 2019-03-20
Add support for fallbacks and relative paths in same way as notmuch

### LINKED ISSUE #232: Database path support does not cover all fallbacks/options
It looks like #219 is tackling support for __$MAILDIR__, which is great!

There are a couple of other ways that notmuch seems to handle the database path. If none is provided, __$MAILDIR__ is used or __$HOME/mail__ as a fallback. ~~Does afew cover the __$HOME/mail__ fallback already?~~ Looks like #219 also covers that.

The next option is to use a relative path. From digging into the latest source (as of March 20th, 2019) of notmuch, it appears to handle relative paths by prefixing them with __$HOME__. Previously, I was running notmuch without a database set and using __$MAILDIR__ and #219 would enable me to use afew; however, I'm looking into switching to providing a relative path so I don't need to define __$MAILDIR__ everywhere (including in cron jobs).

From __notmuch_config.c:670__ of commit __5569e042315862bdedb341472160cd09f4a0a2f7__
Screen Shot 2019-03-20 at 10 16 56 AM

Note that the above logic appears to have been written in 2018-09-08, so it's fairly new. Appears to be part of notmuch 0.28 and later.

Screen Shot 2019-03-20 at 11 20 50 AM

I'll most likely add this support myself in a little while, but wanted to make an issue so I wouldn't forget.

chipsenkbeil:
Resolves #232.

Similar to #219, but also adds support for relative paths in same way as notmuch.

The relative paths is a new feature of notmuch as of 0.28 released back in 2018. Not sure if there's a way to gate support for that by checking the notmuch version.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,13 @@ Support `MAILDIR` as fallback for database location
   supports reading from the `MAILDIR` environment variable, like notmuch CLI
   does, too.
 
+Support relative path for database location
+
+  As of notmuch 0.28, a relative path may be provided for the database
+  location. notmuch prepends `$HOME/` to the relative path. For feature
+  parity, afew now supports the same methodology of prepending `$HOME/` if a
+  relative path is provided.
+
 afew 1.3.0 (2018-02-06)
 =======================
 
--- a/afew/Database.py
+++ b/afew/Database.py
@@ -18,15 +18,23 @@ class Database(object):
     '''
 
     def __init__(self):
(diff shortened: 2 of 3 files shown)


polyzen:
Needs a rebase, and should update `NEWS.md` and `docs/configuration.rst` as in #219.

chipsenkbeil:
Rebased and updated news and configuration.rst

GuillaumeSeren:
Hello,
this is nice patch and quite clean merging thank you !

PR #244, opened 2019-09-21
Implement initial MailMover test suite

kyrias:
I believe that this should cover all possible rule cases and `max_age`.

This is in preparation for another attempt on #147.

### DIFF
--- a/afew/tests/test_mailmover.py
+++ b/afew/tests/test_mailmover.py
@@ -1,8 +1,8 @@
 # SPDX-License-Identifier: ISC
 
-from datetime import timedelta
 import email.message
-from email.utils import format_datetime, localtime, make_msgid
+from email.utils import make_msgid
+from freezegun import freeze_time
 import mailbox
 import notmuch
 import os
@@ -15,10 +15,15 @@ from afew.NotmuchSettings import notmuch_settings, write_notmuch_settings
 
 def create_mail(msg, maildir, notmuch_db, tags, old=False):
     email_message = email.message.EmailMessage()
+    # freezegun doesn't handle time zones properly when generating UNIX
+    # timestamps.  When the local timezone is UTC+2, the generated timestamp
+    # is 2 hours ahead of what it should be.  Due to this we need to make sure
+    # that the dates are always sufficiently far behind 2019-01-30 12:00 to
+    # handle up to UTC+12 .
     if old:
-        email_message['Date'] = format_datetime(localtime() - timedelta(days=20))
+        email_message['Date'] = 'Wed, 10 Jan 2019 13:00:00 +0100'
     else:
-        email_message['Date'] = format_datetime(localtime())
+        email_message['Date'] = 'Wed, 20 Jan 2019 13:00:00 +0100'
     email_message['From'] = 'You <you@example.org>'
     email_message['To'] = 'Me <me@example.com>'
     email_message['Message-ID'] = make_msgid()
@@ -37,6 +42,7 @@ def create_mail(msg, maildir, notmuch_db, tags, old=False):
     return (stripped_msgid, msg)
 
 
+@freeze_time("2019-01-30 12:00:00")
 class TestMailMover(unittest.TestCase):
     def setUp(self):
         self.test_dir = tempfile.mkdtemp()
--- a/afew/tests/test_mailmover.py
+++ b/afew/tests/test_mailmover.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: ISC
 
+from datetime import timedelta
 import email.message
 from email.utils import format_datetime, localtime, make_msgid
 import mailbox
@@ -12,9 +13,12 @@ import unittest
 from afew.Database import Database
 from afew.NotmuchSettings import notmuch_settings, write_notmuch_settings
 
-def create_mail(msg, maildir, notmuch_db, tags):
+def create_mail(msg, maildir, notmuch_db, tags, old=False):
     email_message = email.message.EmailMessage()
-    email_message['Date'] = format_datetime(localtime())
+    if old:
+        email_message['Date'] = format_datetime(localtime() - timedelta(days=20))
+    else:
+        email_message['Date'] = format_datetime(localtime())
     email_message['From'] = 'You <you@example.org>'
     email_message['To'] = 'Me <me@example.com>'
     email_message['Message-ID'] = make_msgid()
@@ -112,6 +116,32 @@ class TestMailMover(unittest.TestCase):
             ])
 
         mover = MailMover.MailMover(quiet=True)
+        mover.move('.inbox', self.rules['.inbox'])
+        mover.move('.archive', self.rules['.archive'])
+        mover.move('.spam', self.rules['.spam'])
+        mover.close()
+
+        with Database() as db:
+            self.assertEqual(expect_inbox, self.get_folder_content(db, '.inbox'))
+            self.assertEqual(expect_archive, self.get_folder_content(db, '.archive'))
+            self.assertEqual(expect_spam, self.get_folder_content(db, '.spam'))
+
+
+    def test_max_age(self):
+        from afew import MailMover
+
+        with Database() as db:
+            expect_inbox = set([
+                create_mail('In inbox, tagged archive, old\n', self.inbox, db, ['archive'], old=True),
+            ])
+
+            expect_archive = set([
+                create_mail('In inbox, tagged archive\n', self.inbox, db, ['archive']),
+            ])
+
+            expect_spam = set([])
+
+        mover = MailMover.MailMover(max_age=15, quiet=True)
         mover.move('.inbox', self.rules['.inbox'])
         mover.move('.archive', self.rules['.archive'])
         mover.move('.spam', self.rules['.spam'])
--- a/afew/tests/test_mailmover.py
+++ b/afew/tests/test_mailmover.py
@@ -1,0 +1,123 @@
+# SPDX-License-Identifier: ISC
+
+import email.message
+from email.utils import format_datetime, localtime, make_msgid
+import mailbox
+import notmuch
+import os
+import shutil
+import tempfile
+import unittest
+
+from afew.Database import Database
+from afew.NotmuchSettings import notmuch_settings, write_notmuch_settings
+
+def create_mail(msg, maildir, notmuch_db, tags):
+    email_message = email.message.EmailMessage()
+    email_message['Date'] = format_datetime(localtime())
+    email_message['From'] = 'You <you@example.org>'
+    email_message['To'] = 'Me <me@example.com>'
+    email_message['Message-ID'] = make_msgid()
+    email_message.set_content(msg)
+
+    maildir_message = mailbox.MaildirMessage(email_message)
+    message_key = maildir.add(maildir_message)
+
+    fname = os.path.join(maildir._path, maildir._lookup(message_key))
+    notmuch_msg = notmuch_db.add_message(fname)
+    for tag in tags:
+        notmuch_msg.add_tag(tag, False)
+
+    # Remove the angle brackets automatically added around the message ID by make_msgid.
+    stripped_msgid = email_message['Message-ID'].strip('<>')
+    return (stripped_msgid, msg)
+
+
+class TestMailMover(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = tempfile.mkdtemp()
+
+        os.environ['MAILDIR'] = self.test_dir
+        os.environ['NOTMUCH_CONFIG'] = os.path.join(self.test_dir, 'notmuch-config')
+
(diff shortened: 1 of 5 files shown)


kyrias:
Hmm, it only fails on Python 3.4 now since it didn't have `run`, 3.4 went end-of-life in March, so does it need to be supported by afew?

flokli:
@kyrias I think only keeping around Python 3.6 and 3.7 are fine.

I openend https://github.com/afewmail/afew/pull/245 for that.

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,152 @@
...
+        (notmuch_msg, status) = notmuch_db.add_message(fname)
+    for tag in tags:
+        notmuch_msg.add_tag(tag, False)
+
+    return (mid, msg)
+
+
+class TestMailMover(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = tempfile.mkdtemp()
```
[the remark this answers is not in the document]
flokli: Using `tempfile` should be enough - they honor `TEMP` if we really want to change the path.

kyrias:
I've rebased and changed so that the mailbox comparisons are done using message IDs rather than filenames, since mids are both more stable, and there's no actual guarantee that the `Maildir` class will use the filenames as keys.  Will squash that into the second commit once approved.

kyrias:
And now it passed on both 3.6 and 3.7.  :)

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+from email.utils import format_datetime, localtime, make_msgid
+import mailbox
+import notmuch
+import os
+import shutil
+import tempfile
+import unittest
+
+
+def get_notmuch_db(db_dir):
```
flokli: isn't that some basic plumbing for which we also have `afew/Database.py`?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+    return notmuch.Database(db_dir, mode=notmuch.Database.MODE.READ_WRITE)
+
+def create_mail(msg, maildir, notmuch_db, tags, old=False):
+    if old:
+        date = format_datetime(localtime() - datetime.timedelta(days=20))
```
flokli: For the sake of reproducibility, can we hardcode some random date in here, instead of using the current time?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+    else:
+        date = format_datetime(localtime())
+    msgid = make_msgid()
+    email_message = email.message_from_string("""\
```
flokli: Can we use `email.message.EmailMessage` to construct the email, instead of assembling the string ourselves?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+To: me <me@example.com>
+Message-ID: {msgid}
+
+{msg}""".format(date=date, msgid=msgid, msg=msg))
+    maildir_message = mailbox.MaildirMessage(email_message)
+    message_key = maildir.add(maildir_message)
+
+    fname = os.path.join(maildir._path, maildir._lookup(message_key))
+    # notmuch 0.26 added index_file and deprecated add_message.
+    if hasattr(notmuch.Database, 'index_file'):
```
flokli: Maybe we can use `afew.Database.add_message`, and add the code there?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+        (notmuch_msg, status) = notmuch_db.index_file(fname)
+    else:
+        (notmuch_msg, status) = notmuch_db.add_message(fname)
+    for tag in tags:
+        notmuch_msg.add_tag(tag, False)
+
+    return (msgid[1:-1], msg)
```
flokli: what are we doing here, and why?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class TestMailMover(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = tempfile.mkdtemp()
+        os.environ['MAILDIR'] = self.test_dir
+
+        self.notmuch_config = os.path.join(self.test_dir, 'notmuch-config')
+        with open(self.notmuch_config, 'w+') as fp:
```
flokli: There's `afew/NotmuchSettings.py` - Maybe we can add this code there?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+            fp.write('[database]\npath={}\n\n[new]\ntags=new'.format(self.test_dir))
+        os.environ['NOTMUCH_CONFIG'] = self.notmuch_config
+
+        # Create notmuch database
+        notmuch.Database(self.test_dir, create=True).close()
```
flokli: use `afew.Database`.

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self.root = mailbox.Maildir(self.test_dir)
+        self.inbox = self.root.add_folder('inbox')
+        self.archive = self.root.add_folder('archive')
+        self.spam = self.root.add_folder('spam')
+
+        self.rules = {
```
flokli: It's not really obvious what this structure does and how we assert on it in the tests below. Can you add some comment and elaborate a bit?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,154 @@
...
+            },
+            '.spam': {
+                'NOT tag:spam AND tag:archive': '.archive',
+                'NOT tag:spam AND NOT tag:archive': '.inbox',
+            },
+        }
+
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
```
flokli: If we're doing everything inside a temporary directory, we can probably use the tempdir contextmanager to let it cleanup for us.

afew/tests/test_mailmover.py
kyrias: There was some reason for not using it originally, but I think I already removed that reason, so fixed.

afew/tests/test_mailmover.py
kyrias: We could, but what would we then pass to `MailMover` as the `max_age` argument?  The only real option would be to then also add the functionality to tell MailMover a fixed date as the `max_age` value.

I'm not sure how this would affect reproducibility though?

afew/tests/test_mailmover.py
kyrias: Stripping the angle brackets off of the message ID that `make_msgid` generated, since the notmuch `get_message_id` function doesn't include them.  I switched to using `strip('<>')` for this instead though.

afew/tests/test_mailmover.py
kyrias: Hm, you mean like some way to pass it a dict and then have that written as the config?  Could do that, though I'm not sure if it would be used by anything other than this one place in the code.

afew/tests/test_mailmover.py
flokli: :+1:

afew/tests/test_mailmover.py
flokli: To diff multiple test run outputs.

Can we use `localtime()` in the test, but use `@freeze_time` from [freezegun](https://pypi.org/project/freezegun/)?

afew/tests/test_mailmover.py
kyrias: How's that comment?

afew/tests/test_mailmover.py
kyrias: That would require each test doing all of the setup as part of the test function instead of using the `unittest` framework's `setUp` function though, but I can do that if you wish I guess?

afew/tests/test_mailmover.py
kyrias: Hm, interesting!  I'll try that out.

kyrias: @flokli There are some issues with this due to the fact that freezegun apparently doesn't handle timezones well.

With the following decorator:

```
@freeze_time("2019-01-30 12:00:00", auto_tick_seconds=15, tz_offset=0)
```

```
now = datetime.now()                 # 2019-01-30 12:00:00
now.strftime('%s')                   # 1548846000
```

And the problem here is that `1548846000` is 12:00 in UTC+1, which is 11:00 UTC.

Meanwhile the date in the mail ends up as `Date: Wed, 30 Jan 2019 13:01:45 +0100` which correctly is 12:00 UTC in UTC+1.  (If switching to `datetime.now()` instead of `localtime()` you get `Date: Wed, 30 Jan 2019 12:01:45 -0000`.)

This means that the date email is an hour ahead of the time returned by `datetime.now()`, which means that it's not included in the timestamp range passed to notmuch.

kyrias: I guess the solution would be to set a static date for the mails created, and it won't be a problem.

kyrias: @flokli 0b37e21 is my attempt at this now.  Is it sufficiently understandable?

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+    def assert_on_expected(self):
+        with Database() as db:
+            in_inbox = { (os.path.basename(msg.get_message_id()), msg.get_part(1).decode())
```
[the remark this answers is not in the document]
flokli: Maybe this can even be a staticmethod not acessing `self` at all, and we remove `assert_on_expected` entirely.

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,162 @@
...
+                'NOT tag:spam AND NOT tag:archive': '.inbox',
+            },
+        }
+
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
+
+
+    def assert_on_expected(self):
```
flokli: I don't really like the way tests mutate `self.expect_*`, and we call that function afterwards. See for a alternative suggestion below.

afew/tests/test_mailmover.py
kyrias: Added `write_notmuch_settings`, and just set `notmuch_settings` instead of doing it in a more roundabout way.

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,162 @@
...
+@freeze_time("2019-01-30 12:00:00")
+class TestMailMover(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = tempfile.mkdtemp()
+        os.environ['MAILDIR'] = self.test_dir
+
+        self.notmuch_config = os.path.join(self.test_dir, 'notmuch-config')
+        with open(self.notmuch_config, 'w+') as fp:
+            fp.write('[database]\npath={}\n\n[new]\ntags=new'.format(self.test_dir))
+        os.environ['NOTMUCH_CONFIG'] = self.notmuch_config
```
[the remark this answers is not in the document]
kyrias: Yes, it being required for `MailMover` is why it needs to be set as an environment variable.

afew/tests/test_mailmover.py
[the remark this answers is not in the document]
kyrias: Added a staticmethod to get the folder content and instead have the `assertEqual` calls within the actual test functions now.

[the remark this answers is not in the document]
flokli: Thanks. This is so much more readable now :)

afew/tests/test_mailmover.py
```diff
@@ -0,0 +1,161 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: ISC
+# Copyright (c) Johannes Löthberg <johannes@kyriasis.com>
```
flokli: Are you fine with dropping the copyright line, for the same reasons as in https://github.com/systemd/systemd/issues/9320:
> Information about who contributed to which file is already covered by `git log` and` git blame` much better. The "copyright by" lines that added manually are often out of date and just unnecessary boilerplate. In particular, all works are covered by copyright without any explicit notice since the 1971 Berne Convention.


I also plan to remove the `# -*- coding: utf-8 -*-` line, as it's a default in Python 3 anyways, so only the `SPDX-License-Identifier: ISC` needs to be preserved.

afew/tests/test_mailmover.py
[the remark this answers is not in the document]
flokli: And it is required due to `MailMover` shelling out to `notmuch` - I don't see anything else using `NOTMUCH_CONFIG` there. Ok :-)

afew/tests/test_mailmover.py
flokli: `# -*- coding: utf-8 -*-` dropped in #247.

kyrias: Done.

PR #245, opened 2019-09-23
drop Python 3.4 and 3.5

flokli:
This allows to use some more recent Python features, and even Debian stable comes with 3.7 by now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 dist: trusty
 language: python
 python:
-  - "3.4"
-  - "3.5"
   - "3.6"
 
 matrix:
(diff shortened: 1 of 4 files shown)


GuillaumeSeren:
We are becoming elitist :snake:

kyrias:
Looks like the tests are only built on 3.7 now: https://travis-ci.org/afewmail/afew/builds/589083000

I think there needs to be a

```yaml
- python: 3.6
  dist: xenial
```

in the `matrix.include sections to run both?

flokli:
@kyrias thanks for catching that! I opened https://github.com/afewmail/afew/pull/246 to fix that.

PR #249, opened 2019-10-01
Abstract mailmover v2

kyrias:
New iteration of #147.


kyrias:
With this implementation, the fully query-based mover ends up as just:

```python
class QueryMailMover(AbstractMailMover):
    def __init__(self, max_age=0, *args, **kwargs):
        super(QueryMailMover, self).__init__(*args, **kwargs)
        self.query = '{subquery}'
        if max_age:
            days = timedelta(int(max_age))
            start = date.today() - days
            now = datetime.now()
            self.query += ' AND {start}..{now}'.format(start=start.strftime('%s'),
                                                       now=now.strftime('%s'))

    def find_matching(self, rule_name, query):
        main_query = self.query.format(subquery=query)
        for message in self.db.do_query(main_query).search_messages():
            for fname in [fname for fname in message.get_filenames()]:
                yield (message, fname)
```

kyrias:
Or, err, not, because apparently it doesn't work yet.

PR #250, opened 2019-10-12
Tree-wide: cleanups

flokli:
This replaces all relative imports, removed unused imports, and reformats some code.

Split into multiple commits for better readability.

### DIFF
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -10,7 +10,7 @@ from afew.Database import Database
 from afew.main import main as inner_main
 from afew.FilterRegistry import all_filters
 from afew.Settings import user_config_dir, get_filter_chain, \
-        get_mail_move_rules, get_mail_move_age, get_mail_move_rename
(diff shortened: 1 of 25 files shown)


afew/commands.py
```diff
@@ -91,13 +89,14 @@
...
+
 def main():
-    if sys.version_info < (3,6):
+    if sys.version_info < (3, 6):
         sys.exit("Python 3.6 or later is required.")
 
     args = parser.parse_args()
 
-    no_actions = len(filter_compat(None, (
+    no_actions = len(filter(None, (
```
kyrias: This doesn't work since Python 3 filter objects do not have a length, so this will therefore raise.  Simplest solution would be to change it to `len(list(filter(...)))`

> TypeError: object of type 'filter' has no len()

flokli: fixed in #255.

PR #255, opened 2019-11-14
Python-3.8 compatibility

jedbrown:
Upon upgrading to python-3.8, I encountered these two issues, which are resolved in this PR.
```
TypeError: object of type 'filter' has no len()
```

```
/home/jed/src/afew/afew/filters/__init__.py:9: SyntaxWarning: "is not" with a literal. Did you mean "!="?                                                                                      
  if filename is not '__init__.py')         
```

### DIFF
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -96,18 +96,18 @@ def main():
 
     args = parser.parse_args()
 
-    no_actions = len(filter(None, (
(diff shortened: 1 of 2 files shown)


tohojo:
Heh, guess I should have looked for this first. Please ignore #256 then :)

PR #256, opened 2019-11-15
Fix crash on Python 3.8

tohojo:
When running on Python 3.8, the following error would appear:

TypeError: object of type 'filter' has no len()

Fix this by calling list() on the result of filter() before passing it to
len().

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>

----

#

### DIFF
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -96,18 +96,18 @@ def main():
 
     args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


tohojo:
Dupe of #255. Sorry for the noise...

PR #266, opened 2020-04-27
Fix trailing ; in notmuch config breaking SentMailsFilter

### LINKED ISSUE #265: (Archive)SentMailsFilter doesn't work if there's an extra ';' in notmuch-config
If there is a trailing semicolon in the notmuch-config the sent mails filter will not work e.g. 
```
primary_email=primary@domain.com
other_email=secondary@domain.com;tertiary@domain.com;
```
This looks like it is because the `other_email` option is handled by 
```python
for other_email in notmuch_settings.get('user', 'other_email').split(';'):
    my_addresses.add(other_email)
```
which (I think) will add an empty string onto the end. Later when the query is built this empty string will result in no emails being matched.

This could be fixed by a simple empty string check in that for loop. However it might be beneficial to adjust the behaviour of `notmuch_settings.get()` to remove any trailing ';' to eliminate any similar bugs.

TimQuelch:
The trailing semicolon resulted in the string split containing an empty string. When the notmuch
query is built later on this results in no messages being matched by the filter.

e.g.  in notmuch-config
```
primary_email=primary@domain.com
other_email=secondary@domain.com;tertiary@domain.com;
```

This fix ensures that the empty string is not added to the `my_addresses` set.

A potentially better fix would be to ensure that `notmuch_settings.get()` removes trailing ;

This fixes #265

### DIFF
--- a/afew/filters/MeFilter.py
+++ b/afew/filters/MeFilter.py
@@ -17,7 +17,7 @@ class MeFilter(Filter):
         my_addresses = set()
         my_addresses.add(notmuch_settings.get('user', 'primary_email'))
         if notmuch_settings.has_option('user', 'other_email'):
-            for other_email in notmuch_settings.get('user', 'other_email').split(';'):
+            for other_email in notmuch_settings.get_list('user', 'other_email'):
                 my_addresses.add(other_email)
 
         self.query = ' OR '.join('to:"%s"' % address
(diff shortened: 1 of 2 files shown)


GuillaumeSeren:
So, I think this patch should take care of all case like that (such in MeFilter).

Ideally the problem should vanish if the list in the notmuch-config did not contain this extra semicolon, but being sure that we don't generate blank item in list can not be wrong.

TimQuelch:
I've changed it to use the existing get_list function. This already filters out empty strings by default.

I also applied the fix to MeFilter, which was the only place that I could find that used the `.get(.*)\.split` pattern.

GuillaumeSeren:
Hey @TimQuelch ,
yes I aggree on the usage of the get_list function as it done on a similar case here https://github.com/afewmail/afew/blob/297e78734c174dd4d2585fb44049fcdb9f75357e/afew/NotmuchSettings.py#L28

I am merging this patch, thank you !

PR #267, opened 2020-05-06
[Bugfix] FolderNameFilter doesn't work on relative paths

BlackLotus:
You can specify as mail root mail/foo instead of /home/username/mail/foo in notmuch, but when doing so afew can't match the folder structure.
My fix checks if it starts with a / (fullpath) and if not adds ~.
An alternative fix would be to expand the matching with ".*", but that would be prone to breaking when your username is the same as the maildir name (e.g. /home/mail/mail)

### DIFF
--- a/afew/filters/FolderNameFilter.py
+++ b/afew/filters/FolderNameFilter.py
@@ -16,6 +16,8 @@ class FolderNameFilter(Filter):
 
         self.__filename_pattern = '{mail_root}/(?P<maildirs>.*)/(cur|new)/[^/]+'.format(
             mail_root=notmuch_settings.get('database', 'path').rstrip('/'))
+        if not(self.__filename_pattern.startswith("/")):
+            self.__filename_pattern = os.path.expanduser("~/"+self.__filename_pattern)
         self.__folder_explicit_list = set(shlex.split(folder_explicit_list))
         self.__folder_blacklist = set(folder_blacklist.split())
         self.__folder_transforms = self.__parse_transforms(folder_transforms)


GuillaumeSeren:
Hello @BlackLotus,
I do not know this use-case, so maybe elaborate on it.

On the patch itself, I think it does work with relative path like '../mail'.

BlackLotus:
Ok let's elaborate a bit :)
afew parses the notmuch config and uses the path used by notmuch, so afew needs to support every path that is supported by notmuch.
For this simple test my mail is located at /home/thomas/mail
```
thomas at archWork in ~/notmyhome
$ ls
nm-1  nm-2  nm-3  nm-4

thomas at archWork in ~/notmyhome
$ cat nm-*|grep "^path"
path=mail/mbsync-rub/
path=~/mail/mbsync-rub/
path=/home/thomas/mail/mbsync-rub/
path=../mail/mbsync-rub/

thomas at archWork in ~/notmyhome
$ notmuch --config=nm-1 count
35400

thomas at archWork in ~/notmyhome
$ notmuch --config=nm-2 count
Error opening database at /home/thomas/~/mail/mbsync-rub//.notmuch: No such file or directory

thomas at archWork in ~/notmyhome
$ notmuch --config=nm-3 count
35400

thomas at archWork in ~/notmyhome
$ notmuch --config=nm-4 count
Error opening database at /home/thomas/../mail/mbsync-rub//.notmuch: No such file or directory
```
So that means that only the fullpatch (/home/thomas/mail/mbsync-rub/) and the relative path to the home directory (mail/mbsync-rub/) are supported by notmuch.

So when I wrote the patch (that yes looks a bit hacky) I checked if the fullpath is used (starts with /) or a relative to the home path.
notmuch can handle '../' in the path name as well, but that doesn't work at the beginning of the path. Here a little bonus test
```
thomas at archWork in ~/notmyhome
$ cat nm-bonus|grep "^path"
path=../thomas/mail/mbsync-rub/.notmuch/

thomas at archWork in ~/notmyhome
$ notmuch --config=nm-bonus count
Error opening database at /home/thomas/../thomas/mail/mbsync-rub/.notmuch//.notmuch: No such file or directory
```
So I think my check covers all possible notmuch paths (at least I hope it does)

BlackLotus:
Oh I just realized I maybe should specify what happens if the patch isn't applied.
https://github.com/afewmail/afew/blob/01e4915935738a87d5106ff78f727d99e94a35e2/afew/filters/FolderNameFilter.py#L30
Tries to match my path structure against the fullpath so for my nm-1 test case this happens:
It tries to match
```
/home/thomas/mail/mbsync.rub/Security Stuff/new/1588640780_4.1572851.archWork,U=6028,FMD5=1597cee2ffabaade9ba69edbe91835ed:2,
mail/offlineimap/(?P<maildirs>.*)/(cur|new)/[^/]+
```
and doesn't find a match.
We could work around that in the regex, but since there are just two cases (full path and path in home) I thought it better to match it exactly instead of having a workaround in the regex that could (potentially) lead to false positives.
(while writing this I realized btw that adding a .* in the beginning will most likely _not_ lead to false positives)

GuillaumeSeren:
Hello @BlackLotus 
thank you for the detailed response.

> afew parses the notmuch config and uses the path used by notmuch, so afew needs to support every path that is supported by notmuch.

I agree on that point, from what I can see on the [notmuch-config man page](https://notmuchmail.org/manpages/notmuch-config-1/),
it support only a top-level directory + a $MAILDIR env variable to prefix.

> So that means that only the fullpatch (/home/thomas/mail/mbsync-rub/) and the relative path to the home directory (mail/mbsync-rub/) are supported by notmuch.

I think what you test show is it support a full path or a top-level path.

You could probably achieve different path with the different MAILDIR env variable,
in each call, and so get afew read different notmuch path each time.

For the FolderNameFilter itself, I think I need more details on what are the path and filename tags,
in theses cases (working nm-1 / nm-3).
[As the doc explain](https://github.com/afewmail/afew/blob/master/docs/filters.rst#foldernamefilter) the message could be in several folders (maildirs) at a given time, so we need to keep that in mind.

```
path: /home/thomas/mail/mbsync.rub/Security
filename: Stuff/new/1588640780_4.1572851.archWork,U=6028,FMD5=1597cee2ffabaade9ba69edbe91835ed:2,
```

This doesn't seems wrong to me, in term of this filter so.

I can see user overriding the regex as an option, instead of changing for all.


ps: You might want to look at :) http://porkmail.org/era/unix/award.html#cat

BlackLotus:
> For the FolderNameFilter itself, I think I need more details on what are the path and filename tags,
in theses cases (working nm-1 / nm-3).
I'm not quite sure which settings you are referencing, sorry. 
My folder transforms look something like
```
folder_transforms = Archiv:archive Archives:archive Security:security Drafts:drafts
```
I could give you a listing of my mail dir structure, but I'm not sure how that would help and I don't think it's what you are asking for :)
> ps: You might want to look at :) http://porkmail.org/era/unix/award.html#cat
Oh no I explicity use cat and grep together because of binary parsing. I could always use grep --text, but I got some corner case in the past that made that not work as expected so I started combining cat and grep again and ignore the use case completely (tbh I don't even remember the case that didn't work with grep --text correctly, but it was 15 years ago or so and now it's a habit).

GuillaumeSeren:
Hello,

> I'm not quite sure which settings you are referencing, sorry.

I am talking new settings, that you can add as parameter of this filter,
that way we can let you define regex for your own use case,
and keep the default for everyone.

I would like to have a summary of what the tag (filename / path) and real path / filename,
to be sure but I am quite convinced that we should not change the default regex.

> PS you are not the first that tries to get me to convert, but old habits die hard and it feels more natural to me when I pre head/tail something

Ah yes I understand that.

I am working on adding test for every filter, so if ever you are motivated to you can help me for this filter, and so we can test both use case of this filter.

BlackLotus:
> I am talking new settings, that you can add as parameter of this filter,
that way we can let you define regex for your own use case,
and keep the default for everyone.

That seems appropriate. I think the fix I committed should not break any setup since it is only used when a non-fullpath is used in notmuch, but not changing the default regex is the smart choice for sure.

 > I am working on adding test for every filter, so if ever you are motivated to you can help me for this filter, and so we can test both use case of this filter.

Seems nice if I ever find the time I will probably take a stab at it, but right now I don' t think I have the time.

GuillaumeSeren:
Hello @BlackLotus,

> I think the fix I committed should not break any setup since it is only used when a non-fullpath is used in notmuch,

Maybe I explained badly,
I don't think we should run this code *at al*l for everyone.

Ideally we add some test for this use-case, and maybe change the default query.

> but not changing the default regex is the smart choice for sure.

What I know, is we rely on notmuch to get this information,
notmuch use a MAILDIR variable if user need to change the prefix,
but changing path in the filter seems confusing (and risky) to me.

> Seems nice if I ever find the time I will probably take a stab at it, but right now I don' t think I have the time.

Yes I understand,
that's why I suggested a *new* user configurable value,
it is simple and change only for the user that want it,
and fast to implement (and document).

BlackLotus:
>  but changing path in the filter seems confusing (and risky) to me.

one of the first things I looked at was how afew read notmuch values and if I should add it there. My thought was that adding a check to an oneliner was the worse choice, but maybe I was wrong.

>  that's why I suggested a new user configurable value,

something that doesn't change backwards compatibility is always a safe, but I don't see the point in adding a configuration option for this particular use case to afew. I made this commit so people that just use a simple afew configuration have a non working setup. If I have to add a specific option to afew I could just as easily change the path in notmuch. The only logical solution is a fix that makes it so that afew works with every possible notmuch value. Either recognizes where the actual maildir is and sets the correct path or change the regular expression to match. I know that risking to break working setups is always something you should avoid, but adding more complexity (for the user, not the codebase) is something I wouldn't really like to see....

GuillaumeSeren:
Hey @BlackLotus

Hum,
let resume why I prefer a user config over this fix.

A user config can used for this use, and probably other,
and did affect every user by default.
As the information should come from notmuch, I don't see why we should changed
that, with an hardcoded value, and maybe this hardcoded value may not be good for
everyone.

With that said,
user should provide a simple way to reproduce the issue,
and I am still not sure about the value of the path / filename returned by the
filter that cause you problem, we need a reproducible case (no code, just
process).

I am trying to help,
please help me keep afew working for everyone *and* solve this issue.

PR #258, opened 2019-12-10
Update quickstart.rst

bjoernfan:
Mention chmod +x for notmuch post-new hook

PR #270, opened 2020-05-27
Improve quickstart

ukleinek:
This is what I happend to notice while looking into afew.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -17,7 +17,7 @@ The following commands will get you going on Debian/Ubuntu systems:
     $ sudo aptitude install notmuch python-notmuch dbacl
     $ git clone git://github.com/teythoon/afew.git
     $ cd afew
-    $ python setup.py install --prefix
+    $ python setup.py install --prefix=~/.local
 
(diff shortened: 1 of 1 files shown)


ukleinek:
Just noticed that this overlaps with #258 .

docs/quickstart.rst
```diff
@@ -51,19 +51,14 @@ Put a list of filters into `~/.config/afew/config`:
...
-
-Then edit the `post-new` file to contain:
-
-.. code-block:: sh
-
-    #!/bin/sh
-    $HOME/.local/bin/afew --tag --new
+    $ notmuchdir=path/to/maildir/.notmuch
+    $ mkdir -p "$notmuchdir/hooks"
+    $ printf > "$notmuchdir/hooks/post-new" '#!/bin/sh\n$HOME/.local/bin/afew --tag --new\n'
```
GuillaumeSeren, line 60: I would prefer a more portable shebang like '#!/usr/bin/env bash',
but this change would require to update everywhere, so it can be done later.

flokli: `/bin/sh` should also exist, and is preferrable if we don't rely on any bashisms.

PR #296, opened 2020-08-28
Add lint action

GuillaumeSeren:
This add a ci action, to lint the source (flake8),
I disabled non-working test so we can see *new* errors.
And fix those _simple_ error in a second time.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,11 +22,15 @@ jobs:
         shell: bash
         run: |
           sudo apt-get update
-          sudo apt-get install -y notmuch python3-notmuch python3-venv
+          sudo apt-get install -y notmuch python3-notmuch python3-venv flake8
           python3 -m venv env
           source ./env/bin/activate
           pip install setuptools pytest dkimpy
           ln -s /usr/lib/python3/dist-packages/notmuch ./env/lib/python*/site-packages
+      - name: flake8 lint
+        run: |
+          source ./env/bin/activate
+          flake8 --ignore=E501 afew/
       - name: Tests
         run: |
           source ./env/bin/activate


flokli:
Can't we just add an additional step to `build.yml`?

Also, I'd prefer if we could fix some of the errors `flake8` complains about, instead of disabling so many.

GuillaumeSeren:
> Can't we just add an additional step to `build.yml`?

Yes I could add this to the build ci, but I think it easyer to read and faster to process.
 
> Also, I'd prefer if we could fix some of the errors `flake8` complains about, instead of disabling so many.

Yes I understand, I will add some more commits to fix those issues.

GuillaumeSeren:
I merged the ci task into the build ci, and started to add fix commit (more to come).

Please have a look and tell me if that seems better to you @flokli

flokli:
> > Can't we just add an additional step to `build.yml`?
> 
> Yes I could add this to the build ci, but I think it easyer to read and faster to process.

This mostly duplicates code from `build.yml`. I don't see how it's more readable.

> > Also, I'd prefer if we could fix some of the errors `flake8` complains about, instead of disabling so many.
> 
> Yes I understand, I will add some more commits to fix those issues.

Thanks. Let's open separate PRs fixing the issues `flake8` complains about, and then merge this PR adding `flake8` to the GitHub action.

I assume some errors types be okay to ignore - let's see what makes sense.

flokli (commented):
please rename the commit messages containing flake fixes to this style:

> tree-wide: fix flake8 E265 - block comment should start with '# '

And open a separate PR for them. It's probably also okay to batch multiple flake8 fixes at once.

GuillaumeSeren:
> And open a separate PR for them. It's probably also okay to batch multiple flake8 fixes at once.

Ok I will work on the fix, and open a seperate PR for them, and after can check to merge this one.

flokli:
With #297 merged, can you rebase this on master (it should then only contain the "add lint step" commit), and remove all the errors from `--ignore` that were adressed in that PR?

GuillaumeSeren:
Hey @flokli ,
I have rebased the commit as you suggested.

One error I kept ignored is the E501 (line too long) because there 83 fix to make,
tell me what you think about it.

flokli (approved):

flokli:
Thanks!

PR #300, opened 2020-10-18
Spam filter to work with flags from Rspamd and bogofilter

### LINKED ISSUE #299: `spamfilter` doesn't work when Rspamd is used instead of SpamAssassin
When Rspamd classifies a mail as spam, it adds the header `X-Spam: Yes` while `spamfilter` searches for the header [`X-Spam-Flag`](https://github.com/afewmail/afew/blob/65b73eb4c80bd0535640330481cb868cd09cf082/afew/filters/SpamFilter.py#L9) which seems to be the flag added by SpamAssassin.

It would be nice to take into account both flags (ie, `X-Spam-Flag` and `X-Spam`).

Titan-C:
I started using bogofilter as it is low on resources, and enough for my email load. I realized I was missing other header matching options for spam. 

This also Closes #299, for Rspamd

P.S. Happy if you also give this PR a hacktoberfest-accepted label

### DIFF
--- a/afew/filters/HeaderMatchingFilter.py
+++ b/afew/filters/HeaderMatchingFilter.py
@@ -22,7 +22,14 @@ class HeaderMatchingFilter(Filter):
     def handle_message(self, message):
         if self.header is not None and self.pattern is not None:
             if not self._tag_blacklist.intersection(message.get_tags()):
-                value = message.get_header(self.header)
+                if not isinstance(self.header, list):
+                    self.header = [self.header]
+
+                try:
+                    value = next(filter(None, map(message.get_header, self.header)))
+                except StopIteration:
+                    return
+
                 match = self.pattern.search(value)
                 if match:
                     sub = (lambda tag:
--- a/afew/filters/SpamFilter.py
+++ b/afew/filters/SpamFilter.py
@@ -6,8 +6,8 @@ from afew.filters.HeaderMatchingFilter import HeaderMatchingFilter
 
 class SpamFilter(HeaderMatchingFilter):
     message = 'Tagging spam messages'
-    header = 'X-Spam-Flag'
-    pattern = 'YES'
+    header = ['X-Spam-Flag', 'X-Spam', 'X-Bogosity']
+    pattern = 'YES|Spam'
 
     def __init__(self, database, tags='+spam', spam_tag=None, **kwargs):
         if spam_tag is not None:


GuillaumeSeren (approved):

GuillaumeSeren:
Hello @Titan-C 
thank you for this patch.

It look's fine to me but this filter does not have test (yet) so I will wait for a second advice to merge it.

nlewo (approved):
Wow, perfect timing (i submitted #299)! Thank you;)

afew/filters/HeaderMatchingFilter.py
```diff
@@ -22,7 +22,14 @@ def __init__(self, database, **kwargs):
     def handle_message(self, message):
         if self.header is not None and self.pattern is not None:
             if not self._tag_blacklist.intersection(message.get_tags()):
-                value = message.get_header(self.header)
+                if not isinstance(self.header, list):
+                    self.header = [self.header]
+
+                try:
+                    value = next(filter(None, map(message.get_header, self.header)))
```
flokli, line 29: This is pretty hard to read. I assume it'll call message.get_header() with every `header` configured in the filter, and filter out results that are an empty string (which can be the case if one of the headers doesn't exist)?

What's the next() in front of everything doing here? Why is `value` still a single element? What happens if the filter returns multiple elements?

flokli (changes requested):
This also needs updates to the documentation, describing the behaviour, and a test (especially considering this changes `HeaderMatchingFilter`, which is used by users directly, and in `ListMailsFilter` too.

afew/filters/HeaderMatchingFilter.py
Titan-C: Yes, you read it right.

`value` is a single element, the first matching header. This is because the `HeaderMatchingFilter` only matches a string, it only has a pattern to match a string. Up until now the filter only matches to one header. For spam, as there are many spam flagging tools, you get different headers depending on the tool, thus you need to be able to search for many headers.
I assume, sysadmins don't run their many spam filters at one, they just pick a tool. Thus I pick the first not empty match(that is what `next` does), which might be the only match. Later I just let the the same logic of the filter run its way and match the pattern to the string. There I use the benefits of the regex. Spamassasing and Rspamd Mark with YES, their spam, bogofilter uses Spam.

flokli: While it makes sense in the scope of `SpamFilter`, this changes semantics in `HeaderMatchingFilter` in a non-intuitive and undocumented way.

Can we expose this in `HeaderMatchingFilter` in a less confusing way, and have it documented?

PR #305, opened 2021-02-03
Handle DMARC report with empty spf or dkim XML nodes

aidecoe:


GuillaumeSeren:
Hey @aidecoe ,
since this filter doesn't have a test yet how do you feel about adding one while working on this issue ?

aidecoe:
Adding another one would be trivial for this case, but to start, it's quite an effort. I needed the quick fix to unblock processing of my mail and so I shared it here. If you don't mind keeping it open for a while, I'll add tests when I get a chance.

GuillaumeSeren:
@aidecoe Yes I know it is an effort but if it can be done it would be so much better.

GuillaumeSeren (approved):

PR #312, opened 2021-10-10
Use watchdog file system watcher to support OSX

benmezger:
I wanted to get the watch feature to work on OSX, so I ended up having to modify and introduce `watchdog`, which supports multiple platforms. Hopefully, someone might find this interesting. At the moment, I am unable to test this on Linux, but it _should work out of the box_, however, if someone could this test on Linux I would appreciate it.

I haven't found a styling guide on this project, so I hope I got the correct coding style.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
           sudo apt-get install -y notmuch python3-notmuch python3-venv flake8
           python3 -m venv env
           source ./env/bin/activate
-          pip install setuptools pytest dkimpy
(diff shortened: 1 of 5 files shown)


benmezger:
Hey @GuillaumeSeren, I've added the corresponding tests.

GuillaumeSeren:
You again @benmezger !
Thank you for sharing this with us.

So this would be a nice addition for mac users, we probably want to build up a small doc / usage / example of this setup.

Beside reviewing I think we have to talk about the storage of the configuration / default values in this patch,
I've read in the other issue that you worked some change in the configuration file maybe some of it could be usefull here.

PR #318, opened 2021-11-19
fix typo in DKIM filter docs

anarcat:
Surely you don't add `dkin-ok` tags (with an `n`)?


GuillaumeSeren:
Hello @anarcat 
Good catch please send more ! :boom:

GuillaumeSeren (approved):

PR #329, opened 2022-08-20
Upgrade python support

GuillaumeSeren:
Hey,
To be able to upgrade python version matrix in the runner:
* Remove python-3.6
* Add python-3.10
* Bump ubuntu to latest stable 20.04
* A few flake8 upgrades

I think it fine, but if it would be better with an other on top of it (if you can @flokli).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
       matrix:
         python-version: ["3.7", "3.8", "3.9", "3.10"]
     name: Build (Python ${{ matrix.python }})
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,7 +30,7 @@ jobs:
       - name: flake8 lint
         run: |
           source ./env/bin/activate
-          flake8 --ignore=E501 afew/
+          flake8 --ignore=E501,W504 afew/
       - name: Tests
         run: |
           source ./env/bin/activate
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,7 +7,7 @@ jobs:
   build-ubuntu:
     strategy:
       matrix:
-        python: [3.6, 3.7, 3.8]
+        python-version: ["3.7", "3.8", "3.9", "3.10"]
     name: Build (Python ${{ matrix.python }})
     runs-on: ubuntu-18.04
     steps:
--- a/afew/files.py
+++ b/afew/files.py
@@ -23,7 +23,7 @@ class EventHandler(pyinotify.ProcessEvent):
         self.database = database
         super().__init__()
 
-    ignore_re = re.compile('(/xapian/.*(base.|tmp)$)|(\.lock$)|(/dovecot)')
+    ignore_re = re.compile(r'(/xapian/.*(base.|tmp)$)|(\.lock$)|(/dovecot)')
 
     def process_IN_DELETE(self, event):
         if self.ignore_re.search(event.pathname):


afew/filters/SentMailsFilter.py
```diff
@@ -21,10 +21,10 @@ def __init__(self, database, sent_tag='', to_transforms=''):
                 my_addresses.add(other_email)
 
         self.query = (
-            '(' +
-            ' OR '.join('from:"%s"' % address for address in my_addresses) +
-            ') AND NOT (' +
-            ' OR '.join('to:"%s"' % address for address in my_addresses) +
+            '(',
```
mjg, line 24: Are you sure you want to change this from a string to a tupel of strings?

GuillaumeSeren: Hey,
I just wanted to fix the that we get on this code:
```
afew/filters/SentMailsFilter.py:25:13: W503 line break before binary operator
afew/filters/SentMailsFilter.py:26:13: W503 line break before binary operator
afew/filters/SentMailsFilter.py:27:13: W503 line break before binary operator
```

But maybe the best is to ignore this warning I am not sure,
what do you think @mjg ?

mjg: Those style checkers can be quite dominating at times ... After all, it's just warnings. For example, exceeding the line width by 1 char is certainly better than making the code less readable.
In this case here: Unless I'm completely wrong, you are switching from string concatenation to a sequence (tupel) of strings. Or did I overlook an additional join or type conversion?

GuillaumeSeren: @mjg thank you for your time.

Yes I agree there is point where we should better ignore this error,
but I am not sur it is the case here, this one is pretty hard to fix. (same for the other change of this commit 40c3bb7d3e61787543a878ddcdf00e93a97de159)

Those 'W504 line break after binary operator' warning is pure style and the fix are not adding anything,
beside cleaning this error:
```diff
diff --git a/afew/files.py b/afew/files.py
index 1cca93e..a0d1415 100644
--- a/afew/files.py
+++ b/afew/files.py
@@ -67,10 +67,7 @@ class EventHandler(pyinotify.ProcessEvent):

 def watch_for_new_files(options, database, paths, daemonize=False):
     wm = pyinotify.WatchManager()
-    mask = (
-        pyinotify.IN_DELETE |
-        pyinotify.IN_MOVED_FROM |
-        pyinotify.IN_MOVED_TO)
+    mask = (pyinotify.IN_DELETE | pyinotify.IN_MOVED_FROM | pyinotify.IN_MOVED_TO)

diff --git a/afew/filters/SentMailsFilter.py b/afew/filters/SentMailsFilter.py
index 78ce3bf..7d57ab2 100644
--- a/afew/filters/SentMailsFilter.py
+++ b/afew/filters/SentMailsFilter.py
@@ -21,10 +21,10 @@ class SentMailsFilter(Filter):
                 my_addresses.add(other_email)

         self.query = (
-            '(' +
-            ' OR '.join('from:"%s"' % address for address in my_addresses) +
-            ') AND NOT (' +
-            ' OR '.join('to:"%s"' % address for address in my_addresses) +
+            '(',
+            ' OR '.join('from:"%s"' % address for address in my_addresses),
+            ') AND NOT (',
+            ' OR '.join('to:"%s"' % address for address in my_addresses),
             ')'
```

I think we can juste drop this commit and this add this warning to ignore,
some related links about it, it look like W504 and W503 are mutully exclusive.
* pylint maintainer suggest to ignore it: https://stackoverflow.com/questions/67942075/w504-line-break-after-binary-operator
* pycodestyle: https://github.com/PyCQA/pycodestyle/issues/792
* packaging: https://github.com/pypa/packaging/issues/144

Considering all this argument I am dropping this and add it to ignore.

PR #336, opened 2023-08-24
HeaderMatchingFilter: Gracefully handle missing headers

tohojo:
Notmuch can raise a NullPointerException when trying to read a non-existent header from a message. Gracefully handle this in the HeaderMatchingFilter by just aborting processing when encountering such an error.

### DIFF
--- a/afew/filters/HeaderMatchingFilter.py
+++ b/afew/filters/HeaderMatchingFilter.py
@@ -6,6 +6,8 @@
 
 from afew.filters.BaseFilter import Filter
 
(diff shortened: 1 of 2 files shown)


GuillaumeSeren:
Hello,
thank you for the patch.

This look like improvement to me but I would like this code to test covered,
do you mind adding a test for this ?

GuillaumeSeren:
Hey @tohojo ,
Thank you for the upgraded patch with the test support for this case.

It is LGTM

GuillaumeSeren (approved):

PR #343, opened 2024-02-28
FolderNameFilter: allow multiple tags per folder

mjg:
So far, the last entry wins when a folder is specified multiple times as a folder transform. The reason is that the folder name is a key in a dict whose values are tags. On the other hand, if a message has copies in several folders then all these rules apply.

Change (fix?) afew's FolderNameFilter so that multiple transforms for the same folder lead to multiple tags being added. Technically, this is done by using sets as values in the transform dict.

In the future, we might think about implementing notmuch-tag style notation (with the obvious quoting issues).

### DIFF
--- a/afew/filters/FolderNameFilter.py
+++ b/afew/filters/FolderNameFilter.py
@@ -54,7 +54,7 @@ class FolderNameFilter(Filter):
         transformations = set()
         for folder in folders:
             if folder in self.__folder_transforms:
-                transformations.add(self.__folder_transforms[folder])
+                transformations.update(self.__folder_transforms[folder])
             else:
                 transformations.add(folder)
         if self.__folder_lowercases:
@@ -71,5 +71,8 @@ class FolderNameFilter(Filter):
         transformations = dict()
         for rule in shlex.split(transformation_description):
             folder, tag = rule.split(':')
-            transformations[folder] = tag
+            try:
+                transformations[folder].add(tag)
+            except KeyError:
+                transformations[folder] = set([tag])
         return transformations


mjg:
FTR: CI runs stone age notmuch (0.29) against py 3.12 which is doomed to fail. The fix for ConfigParser in py 3.12 is in the latest notmuch 0.38 only.

GuillaumeSeren:
Hey @mjg 
Thank you for this patch.

I feel this use case is a bit unclear to me, can you detail a bit how this could append ?

mjg:
> Hey @mjg Thank you for this patch.
> 
> I feel this use case is a bit unclear to me, can you detail a bit how this could append ?

Sure, I didn't meet my own standard on commit messages, sorry :-)

Here's the *why*: Say, I have accounts Work and Pleasure whose mail ends up in subdirs of directories `Work` and `Pleasure` (not that some work couldn't be pleasure, but ...). I want to tag mail by account (for easy display), but also tag sent mail  as `sent` even with clients which just put sent mail in the `Sent` folder. Thus, my afew config is:
```
[FolderNameFilter]
folder_explicit_list = Work/Sent Pleasure/Sent
folder_transforms = Work/Sent:Work Work/Sent:sent Pleasure/Sent:Pleasure Pleasure/Sent:sent
```
and of course more for other folders.
If we had notmuch style config, I would say `Work/Sent:"+Work +sent" Pleasure/Sent:"+Pleasure +sent"`, but we don't, that's why two rules are needed.

I know the account tag search is equivalent to  a path search, but the tag is nice with clients which display it; besides, I can remove/replace the tag and have the `MailMover` move mail between accounts.

GuillaumeSeren (approved):
:+1:

GuillaumeSeren:
Hey @mjg,
Thank you for the detailed explanation, I think it is ok in  the code change,
but there is some bit we better check before merge:
* Do we need to upgrade the docs ?
* Is the existing test cover this case ?

mjg:
> Hey @mjg, Thank you for the detailed explanation, I think it is ok in the code change, but there is some bit we better check before merge:
> 
> * Do we need to upgrade the docs ?

Yes, I should amend `filters.rst`

> * Is the existing test cover this case ?

We don't have any test for `FolderNameFilter` yet, do we? So there's no test I could extend, we would need a new one.

Besides, I would rebase this PR on top of the notmuch2 PR once it is merged. (Merging both works, too.)

GuillaumeSeren:
Hey @mjg 

I merged the #320 PR, I think this one should go next but let's be sure doc and test are as best as we can.

PR #344, opened 2024-04-27
Replace logging.warn usage with logging.warning

jelly:
logging.warn is an alias to logging.warning since Python 3.3 and will be removed in Python 3.13.

https://docs.python.org/3.12/library/logging.html#logging.Logger.warning


GuillaumeSeren (approved):

GuillaumeSeren:
Hey @jelly 

Thank you for the nice patch :+1:

Issue #40, opened 2012-09-03
Configuration is not took in account when using --enable-filters


np:
For instance when using:

```
afew -e ArchiveSentMailsFilter -a
```

I found no way to set the sent_tag option. In particular putting it in the config was useless.

mjg:
`-e` is still not useful, especially with a filter like `FolderNameFilter` which makes no sense without config. I do think that the config should be taken into account.

My use case would be: `notmuch new` does not set any special tag (neither `new` nor `renew`) on renamed messages, so that `afew -t -n` misses out on running `FolderNameFilter` when you move messages to a different folder using IMAP or so. Which, again, defeats the purpose. Running `afew -t -e FolderNameFilter query:myCleverQuery` would be a way to work around.

Alternative approach: patch notmuch...

matclab:
`-e` looks usefull when you want to apply some filters to all mails, (not only `new` one), but you do not want to run all other filters on not `new` mails.

For example, I'm writing a filer to snooze mails, and it had to work on old mail. So my notmuch `post-new` hook looks like:
```
#!/bin/sh
/usr/bin/afew --tag --new
/usr/bin/afew --tag -e SnoozeFilter tag:pending
/usr/bin/afew --move --all
```

varac:
I ran into the same issue. Please fix it, it also took me quite some time to realise that this is indeed a bug and not me.

mjg:
Update: I have a patch for notmuch that tags renamed (moved) messages, I'll try to upstream it shortly.

This would support the mentioned use case: run `afew -e FolderFilter tag:renamed` to adjust the tags. Well, if afew heeds the config, of course.

muradm:
Agreed with @matclab, ability to run single filter without relaying on config file is useful.

Also relates to #237, in the way that, if it would be possible to run `afew` with different configurations, then dropping of `-e` flag won't hurt. If we can use different configs for every execution, that effectively will cover case of `-e`.

PR #347, opened 2024-08-23
Fix the --enable-filters option

drebs:
Filter classes must be initialized with the options found in the config file, but in the case of `--enable-filters` classes were being initialized and config options ignored.

This change makes use of the already present `configured_filter_chain` list of already-initialized filters and trims it down by comparing filter class names with the values passed in the command line.

refs #40


GuillaumeSeren:
Hey @drebs ,
thank you for this patch.

My feeling is if you command override some option you expect it to ignore the config file,
but I may be wrong please elaborate a bit on the use case.

PR #348, opened 2024-09-18
Documentation: Fix malformed quoting in the examples in the configuration doc

marben-olvbar:
When trying to follow the examples in [the configuration documentation](https://afew.readthedocs.io/en/latest/configuration.html#more-filter-examples), I noticed that the queries are quoted, even though they should not be.

I tested this behaviour with Filter.3, which does not produce the expected results due to the quoting.

I assume that the same holds for the others as well. 
If not, please let me know so I can address this.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -146,23 +146,23 @@ Here are a few more example filters from github dotfiles:
 .. code-block:: ini
 
     [Filter.1]
-    query = 'sicsa-students@sicsa.ac.uk'
+    query = sicsa-students@sicsa.ac.uk
     tags = +sicsa
     message = sicsa
 
     [Filter.2]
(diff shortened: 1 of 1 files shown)


GuillaumeSeren:
Hey @marben-olvbar 
thank you for this patch.

I think you are right, in my own config query values are not qutoted,
I want to merge this but can we find some doc justifying this ?

flokli:
The supported format is defined by what `configparser` supports. Their docs only mention:

> This library does not interpret or write the value-type prefixes used in the Windows Registry extended version of INI syntax.

[Wikipedia](https://en.wikipedia.org/wiki/INI_file) notes only some implementations allow quoted values.

Not having wrong examples certainly helps, so let's merge this. I'll push a small edit mentioning values should not be quoted.

PR #352, opened 2025-03-24
Fix GitHub ci node16

GuillaumeSeren:


GuillaumeSeren:
Those are upgrade on the ci that had on issue on the latest python version 3.12+.

* Bump the default target ubuntu version to 24.04
* Add support for python3.13
* Upgrade github action to node 20

.github/workflows/build.yml
```diff
@@ -7,15 +7,15 @@ jobs:
   build-ubuntu:
     strategy:
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
```
flokli, line 10: We can maybe also drop versions that are EOL.

GuillaumeSeren: Yes you are right, I am adding it

NEWS.md
```diff
@@ -1,4 +1,60 @@
-afew 3.1.0 (unreleased)
+afew n.n.n (unreleased)
+=======================
+
```
flokli, line 3: If you look further down in the file, these were usually a bit of a summary, not just a dump of `git log`, which is easier for a user to create on their own. 

In any case, it's unrelated to this PR, so should be in a separate one.

GuillaumeSeren: Ok I'll put that in a seprate PR

PR #363, opened 2025-11-04
Migrate from deprecated pkg_resources to importlib.metadata

Mic92:
The pkg_resources module is deprecated and slated for removal in setuptools 81 (as early as 2025-11-30). This change eliminates UserWarning messages during afew execution by using the modern importlib.metadata API, which has been available since Python 3.8.
Also adds python_requires='>=3.9' to formally declare the minimum version after dropping Python 3.8 support in commit 2a22e7f.


GuillaumeSeren (approved):

GuillaumeSeren:
Hey @Mic92
thank you for the patches

PR #366, opened 2026-02-23
Use updated license specifiers

bkmgit:
Be compliant with PEP 639
https://peps.python.org/pep-0639/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@ setup(
     description="An initial tagging script for notmuch mail",
     url="https://github.com/afewmail/afew",
     license="ISC",
+    license_files=['LICENSE'],
     long_description=long_description,
     long_description_content_type="text/x-rst",
     python_requires='>=3.9',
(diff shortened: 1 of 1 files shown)


flokli:
I didn't find proper documentation describing how this should look like in projects still using `setup.py` - can you point me to the section?

It also seems to be `license` is deprecated, and `license-expression` should be used instead?

bkmgit:
Please see https://setuptools.pypa.io/en/latest/userguide/license_migration.html

flokli:
> Please see https://setuptools.pypa.io/en/latest/userguide/license_migration.html

Sure, but this is all TOML, it doesn't describe how this maps to arguments to call `setuptools.setup` with.

It might make sense to switch to `pyproject.toml` first.

bkmgit:
Ok.

flokli:
4.0.1 has been released, which uses pyproject.toml and the new license specifiers.

PR #368, opened 2026-04-02
NotmuchSettings: add support for profiles and XDG locations

laarmen:
The location ~/.notmuch-config is still supported, but at some point notmuch also grew support for .config/notmuch/$profile/config as well.


flokli:
I'm not a fan of our own config parsing in afew at all.

We can probably use `.config` on a `notmuch2.Database` to read the config, which would automatically follow all config locations.

And the parts writing the config should probably be moved somewhere to the test cases, as they're only use there.

PR #370, opened 2026-07-14
setup.py: exclude tests from wheel but include in sdist

### LINKED ISSUE #367: Don't install test packages
Currently building afew also ships tests.
Test files should be excluded in setup.py

Possibly fix from https://discuss.python.org/t/excluding-a-package-from-a-wheel-install/861/2 
Swap https://github.com/afewmail/afew/blob/23b5aeaa43572a59e95fb00732292087b091d4a1/setup.py#L28 to
```python
    packages=find_packages(exclude=["tests", "tests.*"]).
```

flokli:
Fixes #367

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+graft afew/tests
+global-exclude *.py[cod]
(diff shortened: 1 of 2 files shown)


GuillaumeSeren (approved):

flokli:
I still have some release notes updates from since the last change, push this in a bit

PR #381, opened 2026-07-20
Replace afew/version.py with importlib.metadata.version

wentasah:
Previously, `afew/version.py` was generated by setuptools. Without this file, one gets errors like this:

    Traceback (most recent call last):
      File "/nix/store/mkdw047dmy7ipd1g1pqyrjhp29ldasna-afew-4.0.1/bin/.afew-wrapped", line 6, in <module>
        from afew.commands import main
      File "/nix/store/mkdw047dmy7ipd1g1pqyrjhp29ldasna-afew-4.0.1/lib/python3.14/site-packages/afew/commands.py", line 16, in <module>
        from afew.version import version
    ModuleNotFoundError: No module named 'afew.version'

### DIFF
--- a/afew/commands.py
+++ b/afew/commands.py
@@ -13,10 +13,10 @@ from afew.FilterRegistry import all_filters
 from afew.Settings import user_config_dir, get_filter_chain, \
     get_mail_move_rules, get_mail_move_age, get_mail_move_rename
 from afew.NotmuchSettings import read_notmuch_settings, get_notmuch_new_query
-from afew.version import version
+from importlib.metadata import version
 
 parser = argparse.ArgumentParser()
-parser.add_argument('-V', '--version', action='version', version=version)
+parser.add_argument('-V', '--version', action='version', version=version("afew"))
 
 # the actions
 action_group = parser.add_argument_group(


flokli:
Mmmh, https://setuptools-scm.readthedocs.io/en/latest/usage/ says in their `pyproject.toml` example:

```
# No [tool.setuptools_scm] section needed for basic usage!
```

So I'm a bit baffled this changed when switching from using a `setup.py` to `pyproject.toml` in https://github.com/afewmail/afew/commit/e1fe4344de0ad5abfb39a844fd18a5aa3318e36b.

wentasah:
I'm no Python packaging expert, but I understand that generating of `version.py` is not considered basic usage. As documented [here](https://setuptools-scm.readthedocs.io/en/latest/usage/#version-files), basic usage is perhaps using `importlib.metadata` to access the version info.

mweinelt:
The correct fix is to migrate the version action to use `importlib.metadata.version("afew")`.

wentasah:
Updated the PR based on @mweinelt's suggestion.

mweinelt (approved):

flokli:
Thanks. I also now understand why things initially looked fine locally, I still had a generated `version.py` file in my checkout, which was `.gitignore`d.

I'll remove it from `.gitignore` in a followup, then tag a new release.

